PR #6, opened 2018-04-23
modify setup.py for building windows platform wheels using intel fort…

briantomko:
…ran and visual studio 2015 on python 3.6, disable pyFSQP to get past linker error

Per Justin Gray's request, here is the modifications required to build a platform wheel on windows using the Intel Fortran compiler (for building SNOPT).  I'm not sure if these modifications will break the linux builds.  Perhaps we might want to create a separate setup.windows.py?  If so, reject the pull request and I can go that route.

### DIFF
--- a/pyoptsparse/pyFSQP/setup.py
+++ b/pyoptsparse/pyFSQP/setup.py
@@ -4,16 +4,16 @@ import os,sys
 
 
 def configuration(parent_package='',top_path=None):
-    
+
     from numpy.distutils.misc_util import Configuration
-    
+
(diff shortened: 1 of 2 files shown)


nwu63:
Apologies for not looking into this sooner. Is this still relevant/up to date?

As we don't currently support nor test Windows builds, I'm a little hesitant to include this in the repository. Perhaps some additions to the documentation page is more appropriate? What do you think @briantomko @JustinSGray

JustinSGray:
I don't know if these are current any more. We can close this without merging it for now.

Issue #21, opened 2019-03-04
Implement correct major iteration counting


johnjasa:
We've thought we had it working, but it's not.

Figure out how to determine which iterations were major and save this info in the history database somehow.

eirikurj:
Closing solved by #39

PR #39, opened 2019-06-18
Major iteration counting

nwu63:
Addresses #21 by:
1. calling the SNOPT routine `snkerc` instead of `snoptc`
2. writing a custom `snstop` code in Python to replace the one supplied by SNOPT, and called every major iteration
3. wrapping the `snstop` Python code with f2py and passing it into `snkerc` since `snoptc` does not allow passing a custom `snstop` function handle
4. modifying the db file structure such that the flag `isMajor` is set to true when the evaluation is at a major iteration
5. updating optview accordingly to filter iterations based on this flag


nwu63:
To test, run `test_hs015.py` under `/test`, and examine the resulting `Hist.hst` file both by loading it in Python, and by opening it with optview. The flag `isMajor` should be present and true for major iterations which match the SNOPT print out file. Note that this implementation is able to determine which line search iterations is the actual major iteration, for example `callCounter` 11 is not a major iteration despite being a `funcs` call, and this is evident in the SNOPT print out, where `nCon` is incremented by 2 in major iteration 4.

johnjasa:
@nwu63 what's the status on this? Should I test this out now?

nwu63:
Nobody reviewed/tested the code so never merged. I did bug @eirikurj about it but I know he's been very busy. Would be great if someone can check to make sure it works as expected and then pull it in.

johnjasa:
More changes requested:
With the proposed changes, you can't read older history files because they fail on a KeyError.
Here's a snippet:

```
  File "/home/john/Dropbox/git/pyoptsparse/postprocessing/OptView.py", line 1606, in <module>
    disp = Display(histList, outputDir, figsize)
  File "/home/john/Dropbox/git/pyoptsparse/postprocessing/OptView.py", line 106, in __init__
    self.OptimizationHistory()
  File "/home/john/Dropbox/git/pyoptsparse/postprocessing/OptView.py", line 232, in OptimizationHistory
    self.DetermineMajorIterations(db, OpenMDAO=OpenMDAO)
  File "/home/john/Dropbox/git/pyoptsparse/postprocessing/OptView.py", line 274, in DetermineMajorIterations
    if i == 0 or db[key]['isMajor']:
KeyError: 'isMajor'
```

johnjasa:
We're waiting to merge this in until we have a discussion on the procedure for PyOptSparse updates going forward.

Issue #38, opened 2019-06-18
Prevent SNOPT last call


nwu63:
SNOPT currently does one last function call after performing the whole SQP optimization, with `nState >=2` indicating a final call. I'm not entirely sure what the reason for this is, perhaps to verify the solution? Since our functions are usually very expensive, modify  `_userfg_wrap` as recommended by the manual to do nothing if `nState >= 2`.

nwu63:
Addressed by #43

PR #43, opened 2019-07-31
Fix to skip the final and redundant SNOPT call

nwu63:
## Purpose
Addresses #38 by simply returning without evaluation on the final call.

SNOPT was designed to do a final evaluation after terminating the main loop, so that if you needed to do any final clean up or file IO, this could be handled by your program. Since we don't need this functionality, we can remove the unnecessary (and very expensive!) function call. This call is signified by `nState =>2`, more specifically `nState = 2 + INFO/10` on the final call.
## Type of change

## Testing
SNOPT should now run optimization and exit without the extra function call. For example, with default options the `hs015` case should have the last `callCounter` be 20 instead of 21.

## Checklist

### DIFF
--- a/pyoptsparse/pySNOPT/pySNOPT.py
+++ b/pyoptsparse/pySNOPT/pySNOPT.py
@@ -545,6 +545,12 @@ class SNOPT(Optimizer):
         All we do here is call the generic masterFunc in the baseclass
         which will take care of everything else.
         """
+        # nState >=2 means this is the final call which is redundant
+        # here we just return without doing anything since we don't
+        # need to do any cleanup or anything
+        if nState >= 2:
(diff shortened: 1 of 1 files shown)


johnjasa:
Does the user have the option to have that final function call happen if they set a flag? I forget where we landed on that.

nwu63:
We did say for long term to have options for the user to select what to save in the db file, but didn't touch on that for this particular option. 

Currently the flag doesn't do anything, and this is due to how the `_userfg_wrap` is set up -- it doesn't use the nState parameter at all. In order to make the last call useful, it would be a pretty substantial change, rewriting how the objective function is wrapped and then used in pyoptsparse.

PR #55, opened 2019-10-09
Fix for multiplier post-processing code

gjkennedy:
## Purpose
I fixed the code that post-processes the multipliers after the optimization.

## Type of change

## Testing
I added a test to test_hs071.py to test the multiplier values.

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimization.py
+++ b/pyoptsparse/pyOpt_optimization.py
@@ -1216,7 +1216,8 @@ class Optimization(object):
             else:
                 return fcon
 
-    def deProcessConstraints(self, fcon_in, scaled=True, dtype='d', natural=False):
+    def deProcessConstraints(self, fcon_in, scaled=True, dtype='d',
+                             natural=False, multipliers=False):
         """
         **This function should not need to be called by the user**
 
@@ -1240,6 +1241,11 @@ class Optimization(object):
             Flag to specify if the input data is in the
             natural ordering. This is only used when computing
             gradient automatically with FD/CS.
+
+        multipliers : bool
+            Flag that indicates whether this deprocessing is for the
+            multipliers or the constraint values. In the case of multipliers,
+            no constraint offset should be applied.
             """
 
         if self.dummyConstraint:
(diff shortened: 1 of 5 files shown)


nwu63:
@gjkennedy @JustinSGray  I tried to do the same thing for SLSQP since it also provides Lagrange multipliers, but that runs into 2 problems. 1) there is dimension mismatch for TP109, and 2) the value does not match what you get with SNOPT in HS71. I did verify that the "correct" value in the test is somewhere in the `w` vector in pySLSQP, but I don't know how to go about extracting it. So it seems that the multiplier values in SLSQP are not correct when there is no dimension conflict.

So, if @JustinSGray doesn't need the multipliers in SLSQP then I will just remove `sol.pi` altogether, and only save the multipliers in SNOPT under `lambdaStar` which are scaled. Does that sound good?

JustinSGray:
If possible, I think we should try to get the multipliers from SLSQP. Not everyone has SNOPT, and the multipliers are really valuable for scaling. 

I wouldn't let them hold up progress, but we do have the source for SLSQP right there in the repo. They really should be accessible. no?

gjkennedy:
I've figured out how to extract multipliers from SLSQP. The code that was there was wrong - the variables pi were not the correct values. The multipliers are the first m entries of the w array.

However, the deProecessConstraints function in pyOpt_optimization.py is definitely wrong, and has been this whole time.

nwu63:
Yeah there is some inconsistency with how constraints are counted, linear vs. nonlinear. I don't really know what the variables `optProb.offset` and `optProb.fact` are, so I wasn't really sure how to go about dealing with this. @gkenway and @camader may be able to help?

gjkennedy:
So while, in general, it would be great to have all the optimizers return multipliers, that's clearly beyond the scope of this pull request. It just addresses SNOPT and ParOpt. Others should address the general case right?

jrram:
I agree with Graeme. Better to release a new feature that is only available for a couple of optimizers instead of requiring that all of them be supported (which might never happen).

nwu63:
Okay, merging.

PR #59, opened 2019-11-15
Fixed the printing of the lagrange multipliers.

nbons:
## Purpose
Bug fix for printing of optimization problem/solution.

## Type of change
What types of change is it?

## Testing
Need to add a test that prints the optimization problem so that this doesn't happen again.

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimization.py
+++ b/pyoptsparse/pyOpt_optimization.py
@@ -1607,13 +1607,14 @@ class Optimization(object):
 
         if len(self.constraints) > 0:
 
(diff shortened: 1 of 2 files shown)


eirikurj:
Changes look good. 
For minimal tests I added `print(sol)` to the `test/test_hs015.py` which does print out correctly. This will cover the printing, but we should add more detailed tests in the future.

Kenneth-T-Moore:
This fix didn't fully fix my problem.  I will submit a PR with an updated fix and test shortly.

PR #60, opened 2019-11-15
Fixed a TypeError for unconstrained optimization.

Kenneth-T-Moore:
## Purpose
Recent Lagrange multiplier feature broke unconstrained optimization.  Also made a further fix on the lagrange multiplier printing. 

## Type of change
What types of change is it?

## Testing
Added a test that shows that we no longer get a Type Error exception for an unconstrained problem.

## Checklist


eirikurj:
Ran the new test which passes on py2. 
However, a separate issue was uncovered which has to do with dependency on `six` see, issue #61.

nwu63:
Verified that tests pass on py3, merging.
I'll look into #61 separately.

PR #63, opened 2019-11-20
User termination

Kenneth-T-Moore:
## Purpose
The SNOPT optimizer provides a way to cleanly terminate an optimization by returning a status of -2 or lower from the model function and gradient evaluation functions.  This pull request extends this support through the pyoptsparse interface by allowing the objective and sensitivity callback functions to return a status of 2, which is propagated through pyoptsparse back to SNOPT.

The use-case is defined in the following OpenMDAO enhancement request:
https://github.com/OpenMDAO/POEMs/pull/6

## Type of change
What types of change is it?

## Testing
A test has been added that verifies the feature is working properly.

## Checklist


nwu63:
Everything looks good, and thanks for adding more test coverage. Merging in now.

PR #66, opened 2020-01-02
Added version number to the module

Kenneth-T-Moore:
## Purpose
This PR adds __version__ to the main module so that other packages can check the version of the installed pyoptsparse by querying "pyoptsparse.__version__".  (PEP 396 defines this standard.)

## Type of change
What types of change is it?
- Other (please describe)
This is a fix that further supports pyoptsparse usage as a python package. Adding __version__ allows the version to be queried automatically.  I also updated the version number in setup.py.

Note: These changes make a properly versioned release, so the current 1.1.0 release should be replaced with this commit. You could also choose to increment the release, maybe 1.1.1, but you will have to update the version number in both places.

## Testing
No tests needed.

## Checklist

- [x ] I have run unit and regression tests which pass locally with my changes

### DIFF
--- a/pyoptsparse/__init__.py
+++ b/pyoptsparse/__init__.py
@@ -24,3 +24,5 @@ from .pyALPSO.pyALPSO import ALPSO
 from .pyParOpt import ParOpt
 # from .pyNOMAD.pyNOMAD import NOMAD
 from .sqlitedict.sqlitedict import SqliteDict
+
+__version__ = '1.1.0'
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ if __name__ == '__main__':
 
     setup(
         name             = 'pyoptsparse',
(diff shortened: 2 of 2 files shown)


nwu63:
Apologies in advance -- I'm not too familiar with Python packaging and versioning. PEP 396 [here](https://www.python.org/dev/peps/pep-0396/#classic-distutils) seems to suggest saving the version number in the `init` file, then importing it into `setup.py` so that the version only has to be stored in one file. Would such an approach make more sense than what is done in this PR?

Kenneth-T-Moore:
Sorry I didn't notice the response sooner.  You are right, the way suggested there makes more sense. In fact, we do that in OpenMDAO too.  I've updated the Pull Request to automatically pull the version from the init file so that version can be updated in one place.

PR #71, opened 2020-02-02
Adding travis

nwu63:
## Purpose
PR testing is still not available due to docker image being private, but merging to master should work just fine.

Coveralls is still missing.


nwu63:
This PR should definitely be squashed...

PR #81, opened 2020-03-16
Fixing optimization scaling in solution object

nwu63:
## Purpose
Previously the values stored in the solution object was not scaled back. When scaling is applied, the final solution values change. This PR ensures that the objective, DV, and Lagrange multiplier values stay the same when objective, constraints and DVs are arbitrarily scaled.

Note: the `sol.fStar` field is broken (same as before). I did not have time to debug the issue.

## Type of change
What types of change is it?

## Testing
See the newly-added test `test_snopt_scaling`.

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimizer.py
+++ b/pyoptsparse/pyOpt_optimizer.py
@@ -745,7 +745,7 @@ class Optimizer(object):
         sol.userSensCalls = self.userSensCalls
         sol.interfaceTime = self.interfaceTime - self.userSensTime - self.userObjTime
         sol.optCodeTime = sol.optTime - self.interfaceTime
-        sol.fStar = obj
+        sol.fStar = obj # FIXME: this doesn't work, at least for SNOPT
         n = len(self.optProb.invXScale)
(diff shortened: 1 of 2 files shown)


nwu63:
The value for `fStar` is taken directly from the optimizer, which should work but somehow does not, at least for SNOPT. So instead we access the `optProb.objective` field to retrieve it. I would prefer if we can get it working without that, since it's a bit hacky.

As for the objectives, pyoptsparse was setup with the flexibility to have multiple objectives. However, I don't think any optimizers currently use this. This PR will not scale the Lagrange multipliers by the objective scaling if there are multiple objectives, which I think is a reasonable behaviour.

nwu63:
DVs are scaled correctly. The objective is also scaled correctly _if accessed via the `objectives` field_. The `sol.fStar` field does not have the correct objective.

As for ordering, I personally don't think it's important to maintain consistent argument order when they are keyword arguments. It's going to be difficult to enforce and AFAIK is not part of any PEP. I will however try and get some linting put into the tests on Travis.

PR #83, opened 2020-03-30
readthedocs update

nwu63:
## Purpose
Minor fixes for the documentation to work with readthedocs:
* added the parent folder of `docs` to the `sys.path`, so autodoc can find the modules
* removed current theme, may want to delete the directory in the future, i.e. not ship custom themes with the package

## Type of change
What types of change is it?

- Documentation update

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -16,7 +16,7 @@ import sys, os
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-#sys.path.insert(0, os.path.abspath('../'))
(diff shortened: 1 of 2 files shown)


nwu63:
Our custom theme doesn't render properly on readthedocs for now, and it's quite outdated. I'll bring it up at the meeting, but my thoughts are:
- remove current theme and use the default one for now
- delete the whole theme folder, we shouldn't have that shipped with the package
- if anyone's interested, propose a new theme. This is preferably an existing one which we only have to modify the header/logo, but we can also make major modifications.
- this theme would then be pip-installed also, via the `requirements.txt` file. If it's our own theme, then we would need a repo to host this, and then publish on pyPI. We shouldn't have 10 repos all having the same css files inside, it's absurd.

Issue #74, opened 2020-02-17
Improve pyOpt_History API


nwu63:
## Type of issue
What types of issue is it?
_Select the appropriate type(s) that describe this issue_

- New feature (non-breaking change which adds functionality)

## Description
Add code to pyOpt_History to make it useful for postprocessing. This would include for example
- code to return list of major iterations, and ADflow call counters used for making movies
- code to retrieve various keys at all major iterations (for instance for plotting optimality)
- various other methods for accessing different information within the history file

Ideally optView would be modified to use this API also, instead of doing its own thing

PR #84, opened 2020-04-12
Overhauled history object API

nwu63:
## Purpose
Addresses #74.
The pyOpt_History object has always lacked a full set of API for accessing the history file. As a result, everyone wrote their own scripts to parse through the history file, which causes code duplication. Furthermore, as the history file evolves, certain backwards-incompatible changes have been introduced which will break these codes.

Here we introduce some API for postprocessing an optimization, providing a way to access the data within the history file. In the future, it is hoped that these API will perform version checking against what is stored in pyOptSparse, to provide a stable API for future changes.

## Type of change
What types of change is it?

- Documentation update

## Checklist

### DIFF
--- a/pyoptsparse/pyALPSO/alpso.py
+++ b/pyoptsparse/pyALPSO/alpso.py
@@ -52,7 +52,7 @@ from math import floor
 # External Python modules
 # =============================================================================
 import numpy
-
+from ..pyOpt_error import pyOptSparseWarning
 # =============================================================================
 # Extension modules
 # =============================================================================
@@ -91,7 +91,7 @@ def alpso(dimensions, constraints, neqcons, xtype, x0, xmin, xmax, swarmsize, nh
         if isinstance(x0, list):
             x0 = numpy.array(x0)
         elif not isinstance(x0, numpy.ndarray):
-            print("""Warning: Initial x must be either list or numpy.array,
+            pyOptSparseWarning("""Initial x must be either list or numpy.array,
                 all initial positions randomly generated""")
 
     #
@@ -215,7 +215,7 @@ def alpso(dimensions, constraints, neqcons, xtype, x0, xmin, xmax, swarmsize, nh
 
         else:
             if x0.shape[0] > swarmsize:
-                print('Warning: %d initial positions specified for %d particles, last %d positions ignored'
+                pyOptSparseWarning('%d initial positions specified for %d particles, last %d positions ignored'
                       % (x0.shape[0], swarmsize, x0.shape[0] - swarmsize))
                 x0 = x0[0:swarmsize, :]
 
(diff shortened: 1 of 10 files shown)


nwu63:
One thing I should add is that I did not change the rest of the code base to use the new API. Currently the history file is used in a few places, primarily for the hotstart as well as OptView, but with direct access to sqlitedict.

For the hotstart, I didn't feel comfortable changing it since our test coverage is still not very high, and I want to be absolutely sure that it would not change the functionality because it is a critical feature. For OptView, I don't understand the code there too well and it would take me some time to figure it out, so I've left it out as well. That is definitely a place where others can contribute.

nwu63:
Yeah this feature is definitely long-overdue. I think you raised many good points, the thorough review is appreciated.
- I would like to move away from claiming code ownership via comments in the code (and was planning to purge existing comments like that in the future). Checking the commit history is much more useful and already exists within git.
- I found one docstring that's not complete, will fix.
- The history file is explained in the docs already, see [here](https://pyoptsparse.readthedocs.io/en/latest/api/history.html). Let me know if anything is unclear and I can expand it more. The question about `metadata` should be answered here.
- The point about other optimizers is good. It is the same for all gradient-based optimizers, but stochastic ones like ALPSO and NSGA2 are a bit different there's no `funcsSens`, and evaluations do not encode the major iterations (since it's possible to have many parallel evaluations at each iteration). This part is not really documented so I can explain a bit more.
- The deepcopy has to do with Python always passing variables by reference. If I for example return just `self.conInfo` directly, then the user could modify whatever value they get, and the corresponding instance of `History()` would have its field `History.conInfo` modified. To ensure that the user cannot accidentally modify these values, we return a copy. `copy.copy` vs `copy.deepcopy` is about pointers stored within objects, honestly here `copy.copy` is probably sufficient but I used `deepcopy` just to be safe.
- The scaling feature was added in #15 (which was undocumented!)
- Good call on `deprocessX`. This was added back in #39 for major iteration counting, and is not meant as an API. However, there's some potential code duplication, so I will see if I can remove it.
- I'm not sure I understand what you mean. `getValues` was designed to be the main routine used to retrieve data from history files, and then used for plotting with matplotlib. The `stack` option just puts all DVs in a single key `xuser`, instead of separate keys for different DVGroups. This would be useful for computing for example the L2 distance between different designs.

marcomangano:
Thanks for the clarifications! 
- I did not clearly get the output structure of `getValues` but now I see it is pretty intuitive. Not sure if it is worth adding a bit more comment on that directly on the code or in the docs, even just to indicate that it is the "go-to" function
- I also missed the data structure in the docs page, `metadata` is clear, but I think there still some room for improvement to clarify the different keys. For example, it is not immediately clear why `0` has a `funcs` key and `1` a `funcsens` one. This is a separate documentation question we can discuss elsewhere.

Let us know when you are done with the last minor fixes, so we can merge the PR.

nwu63:
I'm going to add a few examples in the docstring for that. I also noticed that my test coverage is not great which led to at least one bug, so after I fix that I will add more tests. Will finish up tomorrow and let you know when it's ready.

nwu63:
I added a whole bunch of new tests and fixed some bugs. Please take a look and review the code again @marcomangano and @sseraj.

PR #87, opened 2020-04-18
Fix for inconsistent shape in return from getValues

nwu63:
## Purpose
There was a bug in `getValues()` that caused the numpy arrays returned to be inconsistent in shape. This PR removes the messy logic, and simply uses `numpy.stack` to make sure the shapes are consistent. 

This PR also refactored some of the other tests. `assert_almost_equal` from numpy is also used instead of `self.assertAlmostEqual` in Python.

## Type of change
What types of change is it?

## Testing
This issue came up with `names=Hessian`, which is a matrix. When stacked along multiple callCounters, the previous logic fails to return a 3-D array with the correct shape. I have now added tests in `test_hs015.py` to check this.

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_history.py
+++ b/pyoptsparse/pyOpt_history.py
@@ -465,13 +465,14 @@ class History(object):
 
     def getValues(self, names=None, callCounters=None, major=True, scale=False, stack=False):
         """
-        Parses an existing history file and returns a data dictionary used to post-process optimization results.
+        Parses an existing history file and returns a data dictionary used to post-process optimization results, containing the requested optimization iteration history.
 
         Parameters
         ----------
         names : list or str
             the values of interest, can be the name of any DV, objective or constraint,
-            or a list of them. If None, all values are returned.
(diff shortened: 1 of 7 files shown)


nwu63:
Yep exactly, different optimizers are not able to achieve the same tolerance. Part of that has to do with the options -- most of them are set as defaults now and those are inconsistent between optimizers. I'm sure we can tune the options to make them converge tighter, but I think the tolerances are okay for now.

The try/except is there because the optimization problem `hs015` actually has two local minima, denoted by `xStar1` and `xStar2`. In the old code, with take the absolute difference of the solution with respect to both, and enforce that the minimum of the differences is close to 0. In the new code, we just assert twice, with try/except so that if it's not near the first optimum we assert that it must be near the second. The use of `assert_all_close` also makes sure we're taking the appropriate norm for vector quantities -- previous we use `np.min` so we're not really checking all components.

Actually, reading the numpy documentation further they seem to recommend using `assert_allclose` instead, so I will change that and then push my changes up.

Issue #90, opened 2020-04-20
pyIPOPT doesn't return the IPOPT return code


robfalck:
## Type of issue
What types of issue is it?

- Bugfix (non-breaking change which fixes an issue)\

## Description

pyIPOPT contains a `solInform` property but it is not set to the appropriate return code from IPOPT.

### Steps to reproduce issue

The hs_071 IPOPT test has been modified to test this behavior:

```
    def test_ipopt(self):
        opts = {}
        opts['print_level'] = 5
        opts['linear_solver'] = 'mumps'
        sol = self.optimize('ipopt', optOptions=opts)
        self.assertEqual(sol.optInform['value'], 0)
        self.assertEqual(sol.optInform['text'], 'Solve Succeeded')
        # Test that the inform is -1 when iterations are too limited.
        opts['max_iter'] = 1
        sol = self.optimize('ipopt', optOptions=opts, check_solution=False)
        self.assertEqual(sol.optInform['value'], -1)
        self.assertEqual(sol.optInform['text'], 'Maximum Iterations Exceeded')
        # Test that the inform is -4 when max_cpu_time are too limited.
        opts['max_iter'] = 100
        opts['max_cpu_time'] = 0.001
        sol = self.optimize('ipopt', optOptions=opts, check_solution=False)
        self.assertEqual(sol.optInform['value'], -4)
        self.assertEqual(sol.optInform['text'], 'Maximum CpuTime Exceeded')
```

### Current behavior

Currently sol.optInform provides an empty dictionary for optInform.

### Expected behavior

optInform should contain the return code provided by IPOPT and the corresponding message.  These are available from https://github.com/coin-or/Ipopt/blob/master/src/Interfaces/IpReturnCodes_inc.h

### Code version (if relevant)

N/A

**External dependencies:**

N/A

**Internal packages:**

N/A

PR #91, opened 2020-04-20
Ipopt informs

robfalck:
## Purpose

The current version of pyIPOPT.py was not returning the return code from IPOPT.
The file has been updated to handle the return codes as given in https://github.com/coin-or/Ipopt/blob/master/src/Interfaces/IpReturnCodes_inc.h

Resolves: #90 

This is a fix for master but the same changes apply to v1.2 (for those people without the latest SNOPT).  They are available on branch ipopt_informs_v1_2 on my fork. (I don't seem to be able to submit a PR against a fork).

## Type of change

## Testing

I have tested this fix locally on my system and it works as expected.  I have added tests of the inform value to IPOPT tests in the code.  All that is needed is to test that `solInform` contains the appropriate number and message.

## Checklist

### DIFF
--- a/pyoptsparse/pyIPOPT/pyIPOPT.py
+++ b/pyoptsparse/pyIPOPT/pyIPOPT.py
@@ -30,7 +30,7 @@ except ImportError:
 # standard Python modules
 # =============================================================================
 import copy
-import time, datetime
+import time
 # =============================================================================
 # External Python modules
 # =============================================================================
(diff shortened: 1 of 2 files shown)


nwu63:
Hi Rob, everything looks good here. Just a question: what version of IPOPT did you run this with? AFAIK the latest version we support is several years out of date, so I'm curious to know if you got it working with a more recent version.

robfalck:
Good question!  One of our OpenMDAO devs (@tadkollar) wrote a script that downloads the appropriate version of IPOPT, its available at this gist: https://gist.github.com/robfalck/9fcefe17464bd1e6dccf0ccb21e59f46

It's a bash script, but we'd love it if this could be made part of the setup.py for pyoptsparse (getting IPOPT to work is not trivial).

nwu63:
Huh, I was under the impression that the pyIPOPT wrapper we ship with pyoptsparse is not compatible with newer versions of IPOPT, and quite a bit of work is needed to get pyoptsparse to work with that. But if what you're saying is the case then that is really good to know -- we can at least try it out and update the docs/instructions to support the newer versions.

As for installing, I'm really not that familiar with `setup.py`. To be honest I'm more inclined to have it installed as a separate package, similar to what we do currently with `ParOpt`, rather than copy the files over to within `pyOptSparse` and have its `setup.py` handle it. Thanks for the bash script, that's definitely going to be helpful. I haven't played around with IPOPT for some time but it's something I actually plan on picking up again relatively soon.

robfalck:
I'd be OK with that. Anything that makes installing the full stack easier is a good thing.  We understand that script works well for our team but might not be ideal for everyone.

nwu63:
We're in the same spot. As we move towards using docker containers we hope to provide users with images containing open-source optimizers pre-installed, and make the script available for those that want to install locally, similar to the gist you've shared.

PR #92, opened 2020-04-21
Code cleanup for python 3

nwu63:
## Purpose
This PR does the following:
- Update Travis to test using our new py3 containers
- Remove old code for py2.6 compatibility
- Remove future import statements
- Remove Author/History information in module docstrings

## Type of change
What types of change is it?

## Testing

## Checklist


nwu63:
The centOS test is failing due to the use of an old compiler for IPOPT, which will be fixed relatively soon.

nwu63:
Yeah I think it's fine since there is not much code change in this PR. Don't think you can merge without admin rights, so if both reviewers approve this today then I will merge it without waiting for new Docker images.

PR #93, opened 2020-04-22
Refactoring the scaling code

nwu63:
## Purpose
This PR refactors the mapping between values in the user-space, and values in the optimizer-space. We allow `x' = (x - offset) * scale` type transformations for the DVs, and `val' = val * scale` type transformation for the objective and constraint values. Previously this was done in an ad-hoc fashion whenever the mapping is needed. I've refactored the code, and added various `_map` functions to do this more explicitly and consistently. I've gone through the code and made sure that all transformations are done via these `_map` functions.

There are some other refactoring:
- Removed the `optProb.bulk` attribute since it is not used
- The `optProb` is also now included in the History file. We have to remove the `comm` and `objFun` attributes to make the class instance serializable
- Renamed the `process` and `deProcess` functions to be more clear. Also added `processObjtoDict` which was missing
- Added dictionary-based mapping functions`_Dict`
- Updated the History class to use these new functions for mapping
- Added new tests for optimization with `xOffset`, and tests for scaling/mapping the optimization results.

## Type of change
What types of change is it?

- Refactoring (no functional changes, no API changes)

## Testing
I've added new tests:
- `TestHS71.test_NLPQLP_setDV`
- `TestHS71.test_snopt_setDVFromHist`
- `TestHS71.test_slsqp_scaling_offset_optProb`

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,13 +14,17 @@ env:
     - NLPQLP_DIR=$DOCKER_WORKING_DIR/pyoptsparse/pyNLPQLP/source
     - IPOPT_DIR=$DOCKER_WORKING_DIR/pyoptsparse/pyIPOPT/Ipopt
   jobs:
-    - DOCKER_TAG=py2
+    - DOCKER_TAG=py2-latest
+    - DOCKER_TAG=py3-stable
+    - DOCKER_TAG=py3-latest
+    - DOCKER_TAG=py3-impi
+    - DOCKER_TAG=py3-centos7-impi
 
 before_install:
   - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
       export DOCKER_REPO=public;
     else
-      export DOCKER_REPO=opt;
+      export DOCKER_REPO=private;
       echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin;
     fi
   - docker pull mdolab/$DOCKER_REPO:$DOCKER_TAG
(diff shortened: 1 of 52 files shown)


nwu63:
Hmm I did not anticipate problems with python 2, stand by.

nwu63:
Good point, I am going to remove `_deProcessX` and instead use the `optProb.deProcessX`, since that is now available. In the future, I will think of a better way to organize the optimization information. For example, `conInfo` and `optProb` contain duplicate information, but in slightly different formats.

The `finalize` functions do more than that. They loop through all the DVs/constraints, and construct various parameters which are used later on, such as scaling arrays.

nwu63:
Yes really good point, I will improve the documentation. I am also planning to add some (long-overdue) unit tests --- these mapping functions for example would be a perfect candidate for unit testing.

Sorry about the mess, I didn't realize this was going to get this big, and I should have submitted the PR only when everything is ready. Will let you know when it's ready for review, and thanks for taking the effort to look through this!

nwu63:
@marcomangano I've added the unit tests I promised, and added some more docstrings. Please take a look when you've got time.

nwu63:
The call ``opt(self.optProb, 'FD')`` will actually run the optimization. This is needed because certain things within the ``optProb`` are only populated once you call the optimizer. This includes optimizer-specific ordering of constraints, for example. Ideally that is done in a pre-processing or finalization step, and refactored, but it is not done. I do plan to add additional tests though, maybe I will open an issue to keep track of what we should test.

I've added most of the docstrings you mentioned. I didn't add anything for the ``_map`` functions though, since they're private member functions and I felt they were pretty self-explanatory (there's also a lot of them to document).

I do want to do some more refactoring and type checking, I feel there needs to be more rigor in what happens in pyOptSparse, for example when casting to complex or real, or when casting scalars to numpy arrays with a single element. This is an inherent problem with Python that we have to be careful about.

PR #94, opened 2020-04-25
Updated coveralls setting

nwu63:
## Purpose
I updated travis and coveralls settings, to omit certain files when calculating the code coverage.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,10 +7,10 @@ exclude_lines =
     if __name__ == .__main__.:
 
 [run]
-
 omit =
     */build/*
     */doc/*
     */examples/*
     */postprocessing/*
     */test/*
+    */setup.py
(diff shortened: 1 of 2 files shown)


nwu63:
- `NOMAD` is not installed by default (not sure why actually) but it's commented out [here](https://github.com/mdolab/pyoptsparse/blob/master/pyoptsparse/setup.py#L22)
- `FSQP` does not build
- `NLPY_AUGLAG` source code is not part of pyoptsparse, and I couldn't find a place to find it actually
- `ALPSO_ext` is only used for parallel execution as far as I understand

So none of these are being tested. Perhaps we should talk about whether we want to still ship/support these optimizers. IMO we don't need to host tons of old code in Fortran that's not being maintained, and I don't think anyone is using any of these with the exception of `ALPSO`, which I would say is poorly supported.

PR #99, opened 2020-04-29
Fixed bug where getValues() tries to retrieve linear constraint values

nwu63:
## Purpose
`getValues` tries to loop over all the `conNames`, which contains all the linear constraints too. So, when constructing the dictionary of values, it then tries to read those linear constraint values from `funcs`, which are of course not stored there. So I've modified it to just only read from `funcs` if the constraint is nonlinear.

## Type of change
What types of change is it?

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_history.py
+++ b/pyoptsparse/pyOpt_history.py
@@ -436,7 +436,9 @@ class History(object):
         """
         conDict = {}
         for con in self.conNames:
-            conDict[con] = d['funcs'][con]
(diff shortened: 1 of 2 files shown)


nwu63:
Really glad that a test caught this, I just pushed up a fix.

@Xiaosong2105, the linear constraints are stored in `optProb.constraints`, same as the other constraints. But since they are linear, the Jacobian is constant and there is no need to evaluate them in `funcs` or `sens`. Instead the Jacobian is stored explicitly in the constraint object.

PR #100, opened 2020-05-03
Apply formatting and flake8 fixes

nwu63:
## Purpose
I've formatted the code via `black`, and fixed all the errors with `flake8`. I also changed `numpy` to `np`, and incremented the package version in anticipation of a new minor release.

I plan to use this PR to also test a new CI setup which will check `black` and `flake8` compliance on PRs. **Please do not merge this until that is good to go.**

## Type of change
What types of change is it?

## Checklist


nwu63:
Sorry for the messy commits, but I think that was the fastest way to sort this stuff out. I have now added 4 GitHub Actions, one for `black` and three for `flake8`, with different python versions. Everything seems to work well so I think the PR is ready. If others feel this is a good way to do this, then we can have this set up in our other repositories also.

marcomangano:
The checks are really quick so I don't see any issues! In case the codestyle checks don't pass is it still possible to merge? And is there a way to fix the code on the fly or is it best to fix the PR locally and merge again as you did?
We should also draft a schedule (and share with the lab) about the conversion of all the other repos.
I think this PR is ready to be merged anyway!!

nwu63:
I haven't added them to the mandatory checks yet, but that's the plan. No the PR author should fix everything locally, then push up another commit.

For the other repositories, it may be too much to enforce flake8 from the beginning, so perhaps the check should not be made mandatory. But `black` should definitely be enforced. It's also possible to let flake8 only check the commit diff, i.e. only the parts that the PR changed, rather than the entire repo.

marcomangano:
This last option sounds like a good way to facilitate transition! Let's talk about this separately

PR #103, opened 2020-05-05
Fix to multiplier extraction code/update to most recent ParOpt

gjkennedy:
## Purpose
This pull request fixes the multiplier extraction for "oneSided" optimizers. pyOptSparse handles these optimizers by adding two versions of the same constraint for the lower and upper bound. Before, the processContoDict code did not properly select the multiplier estimate associated with the active constraint and applied the scale factors incorrectly.

## Type of change
What types of change is it?

## Testing
This code was tested using ParOpt, which seems to be the only optimizer other than SNOPT that provides multipliers at the solution. The new code should work by running any of the examples, for example:

python tp109.py --opt ParOpt

## Checklist


pyoptsparse/pyOpt_optimization.py
```diff
@@ -1274,18 +1274,23 @@ def processContoDict(self, fcon_in, scaled=True, dtype="d", natural=False, multi
...
-
         # Perform constraint scaling
         if scaled:
-            fcon_in = self._mapContoOpt(fcon_in)
+            m = len(self.jacIndices)
+            fcon_in[:m] = fcon_in[:m]*self.conScale[self.jacIndices]
+
+        fcon_unique = fcon_in
+        if multipliers:
+            fcon_unique = numpy.zeros(self.nCon)
```
sseraj, line 1284: It looks like your last commit reverted np to numpy.

pyoptsparse/pyParOpt/ParOpt.py
```diff
@@ -382,6 +389,10 @@ def _set_paropt_options(self, opt):
                     opt.setPenaltyDescentFraction(value)
                 elif key == "min_penalty_parameter":
                     opt.setPenaltyDescentFraction(value)
+                elif key == "min_penalty_parameter":
```
sseraj, line 392: This elif is a duplicate of the previous one.

pyoptsparse/pyParOpt/ParOpt.py
```diff
@@ -223,32 +240,20 @@ def evalObjConGradient(self, x, g, A):
...
-                opt.setOutputFile(filename)
-
-                # Set the penalty parameter internally in the
-                # code. These must be consistent between the trust
-                # region object and ParOpt.
-                opt.setPenaltyGamma(tr_penalty_gamma)
-
-                # Set parameters for ParOpt in the subproblem
-                opt.setMaxMajorIterations(tr_max_iterations)
-                opt.setAbsOptimalityTol(tr_opt_abs_tol)
```
nwu63, line 246: @gjkennedy does this function call no longer exist in the new version of ParOpt? `flake8` is complaining that `tr_opt_abs_tol` is now no longer used.

gjkennedy:
Thanks!


--
Graeme J. Kennedy, PhD
Assistant Professor
School of Aerospace Engineering
Georgia Institute of Technology
270 Ferst Dr, Atlanta, GA 30332-0150
http://gkennedy.gatech.edu<http://gkennedy.gatech.edu/>


________________________________

pyoptsparse/pyParOpt/ParOpt.py
```diff
@@ -257,28 +262,28 @@ def evalObjConGradient(self, x, g, A):
                 else:
                     opt.setNormType(_ParOpt.L2_NORM)
 
-                # Initialize the problem
-                tr.initialize()
+                # Set the ParOpt options
+                self._set_paropt_options(opt)
```
nwu63, line 266: Is this call necessary, or a duplicate of line 255? And since the `setNormType` call is already in the `set_paropt_options` call, do they have to be set again here?

nwu63: @gjkennedy can you take a look at this when you've got a chance? And the same with the `tr_opt_abs_tol` value above.

gjkennedy: The norm_type is set in set_paropt_options so it can be removed here.

pyoptsparse/pyParOpt/ParOpt.py
gjkennedy: opt.setAbsOptimalityTol(tr_opt_abs_tol) should be called after set_paropt_options so that the tolerance that the trust region subproblems are solved to can be set independently from the overall optimization problem.

nwu63: Right, there is no `opt.setAbsOptimalityTol(tr_opt_abs_tol)` call at the moment.

pyoptsparse/pyParOpt/ParOpt.py
nwu63: Yeah I agree. Also, `_set_paropt_options` is called twice, on L255 and L266, which I think is redundant, right?

nwu63:
For the sake of not letting this PR sit for too long, I will merge this and then open another PR for you to review @gjkennedy.

Issue #105, opened 2020-05-19
pyoptsparse can be installed without sqlitedict being installed


bbrelje:
## Type of issue
What types of issue is it?
- Other (please describe)
Installer update

### Steps to reproduce issue
1. Install pyoptsparse per the documentation
2. Installation is apparently successful but sqlitedict dependency is not enforced
3. Errors appear downstream e.g. https://github.com/OpenMDAO/OpenMDAO/issues/1422

nwu63:
SqliteDict has been moved to an external dependency, this was described both in the release notes for v1.2, and in the documentation, so that's an expected behaviour. To be clear, pyOptSparse does not do dependency checking during install. However, we are due for an overhaul of the install process, so I'll leave this up as a feature request.

bbrelje:
I don't understand what you mean by "moved to an external dependency". Do you mean that pyoptsparse used to provide a home-grown implementation of sqlitedict?

setuptools style is to enforce dependencies which the project requires to "minimally" run (this qualifies) but not pin versions
https://packaging.python.org/discussions/install-requires-vs-requirements/

nwu63:
Yes pyOptSparse used to ship with an older version of sqlitedict, and the switch has caused problems for some people.

I understand, we currently provide a `requirements.txt` file to help with that, but do not enforce anything during install. I don't have much experience with Python installs, but if this is a simple change to the `setup.py` script then would you mind doing a PR for that?

bbrelje:
Done

PR #106, opened 2020-05-19
require sqlitedict for minimal installation closes 105

bbrelje:
## Purpose
Do not allow installation via pip or setuptools without required dependencies installed

## Type of change
What types of change is it?

## Testing

uninstalled sqlitedict
installed per docs
sqlitedict installed automatically

## Checklist
- [ ] I have run unit and regression tests which pass locally with my changes (N/A)
- [ ] I have added new tests that prove my fix is effective or that my feature works (N/A)
- [ ] I have added necessary documentation (N/A)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,6 +17,7 @@ jobs:
   flake8:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: [ '2.7', '3.7', '3.8' ]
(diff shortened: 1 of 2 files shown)


bbrelje:
Closes #105

nwu63:
I checked this in a Docker container, and it works as expected for pip installations. However this does not work when running `setup.py` manually. Should we just remove that installation option?

Also, we may want to remove `requirements.txt` since that information is encapsulated in the `setup.py` file now.

PR #108, opened 2020-05-26
Minor format fix

nwu63:
## Purpose
This is a minor formatting fix. Due to the way things were coded before, long error messages were concatenated by just a series of strings. When I reformatted with black, it combined certain lines which were not long enough into a single line, resulting in these extra quotes. I removed them in this PR.

## Type of change
What types of change is it?

### DIFF
--- a/pyoptsparse/pyALPSO/alpso_ext.py
+++ b/pyoptsparse/pyALPSO/alpso_ext.py
@@ -71,7 +71,7 @@ def alpso(dimensions, constraints, neqcons, xtype, x0, xmin, xmax, swarmsize, nh
             x0 = np.array(x0)
         elif not isinstance(x0, np.ndarray):
             pyOptSparseWarning(
-                ("Initial x must be either list or numpy.array, " "all initial positions randomly generated")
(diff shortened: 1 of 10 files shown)


nwu63:
First of all, keep in mind that the code containing `" "` was not _incorrect_. It's ugly, but it works fine.

The issue is not with black, but with the original code style. See for example line 80-81 [here](https://github.com/mdolab/pyoptsparse/blob/v2.0.3/pyoptsparse/pyNLPQLP/pyNLPQLP.py#L80). The string was just joined implicitly with a backslash.

It's fairly well documented that this is bad, see the explanation [here](https://stackoverflow.com/a/2059025). I personally prefer explicit string concatenation, i.e. via the `+` operator. Whether or not you surround it in quotes is a personal preference I guess.

nwu63:
Just a side note: `psf/black` recently added [better string handling](https://github.com/psf/black/pull/1132), so it's possible that with this new feature, it could do this more intelligently. The PR was merged but it has not made its way into a stable release yet, so I haven't tested it.

PR #112, opened 2020-05-29
Removing support for paropt

nwu63:
## Purpose
We have decided to temporarily remove support for ParOpt, meaning we will no longer build ParOpt in our Docker images, and run pyParOpt tests within pyOptSparse.

This PR updates the documentation to reflect that change.

## Type of change
What types of change is it?

- Documentation update


nwu63:
Yep we didn't change anything in this repo, but we have disabled ParOpt tests both for the nightly build, and for Travis. Re-enabling them will be easy to do.

PR #115, opened 2020-05-31
Set math rendering to mathjax and updating date

eirikurj:
## Purpose
- Set math to be rendered as mathjax. This offers the possibility of copying expressions over to latex.
- Update the year

## Type of change
- Documentation update

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -31,6 +31,7 @@ extensions = [
     "sphinx.ext.viewcode",
     "numpydoc",
     "sphinx.ext.autosummary",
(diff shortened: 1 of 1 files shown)


nwu63:
I believe MathJax is injected by readthedocs automatically. This approach is far more explicit, but you are correct there is no change to behaviour.

nwu63:
There is math stuff [here](https://mdolab-pyoptsparse.readthedocs-hosted.com/en/latest/guide.html), granted there is not a lot.

PR #116, opened 2020-05-31
Updated conf.py

nwu63:
## Purpose
The `conf.py` file used is autogenerated but from a very old version of sphinx. It contained many entries which are not useful -- I have autogenerated a new version and copied over useful entries.


nwu63:
1. The reason we specify them for the Optimization class is because it has a lot of private member functions. We _could_ convert them to actual private functions by prepending them with an underscore, but I didn't want to do that since that's a much bigger change. Eventually, I do think we should do that though.
2. We still have to figure out how to do the release version. We shouldn't have to specify it here otherwise we'd have to update this whenever we release a new version. I'll look into it.

Issue #117, opened 2020-06-01
Provide a way to let a user tell pyoptsparse to NOT use mpi4py


hschilling:
## Type of issue
What types of issue is it?
_Select the appropriate type(s) that describe this issue_

- New feature (non-breaking change which adds functionality)

## Description
Provide a way to let a user tell pyoptsparse to NOT use mpi4py.

I work on the OpenMDAO team and it has happened that sometimes we can totally break our MPI installation on our dev machines. It's handy to tell OpenMDAO to skip importing mpi4py if just want to run some code that doesn't make use of MPI while you work on fixing the MPI install.

So there is an environment variable, OPENMDAO_REQUIRE_MPI, that can be set to '0'. (Really anything other than ['always', '1', 'true', 'yes']) and then OpenMDAO skips importing mpi4py.

The problem is that we also obviously import pyoptsparse and it imports mpi4py no matter what.  We have no control over that. 

I see two places in the pyoptsparse code where code would need to be added to give the user control:

- At the bottom of pyOpt_MPI.py
- At the top of pyParOpt/ParOpt.py 

### Current behavior
When using pyoptsparse, there is no way to force pyoptsparse to skip importing mpi4py.

### Expected behavior
Provide some environment variable ( or some other mechanism ) for the user to set so that mpi4py is not imported. 

### Code version (if relevant)
NA

nwu63:
Thanks for writing this up, I can see why you guys want this feature. In my mind this is relatively straightforward to implement -- wrap the MPI imports with a conditional based on some environment variable. We could define something like `PYOPTSPARSE_REQUIRE_MPI` for this purpose (and then document this behaviour of course). Note that `ParOpt` requires MPI, so it will have to be disabled if `mpi4py` is not used.

Would you/others on the OpenMDAO dev team be willing to do this and then make a PR?

hschilling:
Yes, I would be happy to. FYI, here is an example of how we do it in OpenMDAO:

```# Attempt to import mpi4py.
# If OPENMDAO_REQUIRE_MPI is set to a recognized positive value, attempt import
# and raise exception on failure. If set to anything else, no import is attempted.
if 'OPENMDAO_REQUIRE_MPI' in os.environ:
    if os.environ['OPENMDAO_REQUIRE_MPI'].lower() in ['always', '1', 'true', 'yes']:
        from mpi4py import MPI
    else:
        MPI = None
[...]

PR #118, opened 2020-06-02
Added control over using mpi4py by setting environment variable PYOPTSPARSE_REQUIRE_MPI

hschilling:
## Purpose
This PR fixes issue #117.

The goal is to give the user some control over using mpi4py. The user can either say they don't want to run in parallel or explicitly say they must run in parallel. Otherwise, pyoptsparse currently tries to run in parallel as needed but if a suitable mpi4py is not available, there is a fallback to run in serial.

## Type of change
What types of change is it?

## Testing
One test is to set `PYOPTSPARSE_REQUIRE_MPI` to '0'. Any tests that use ParOpt should fail since ParOpt needs to run in parallel and the environment variable is telling pyoptsparse to run in serial. 

## Checklist

### DIFF
--- a/pyoptsparse/pyParOpt/ParOpt.py
+++ b/pyoptsparse/pyParOpt/ParOpt.py
@@ -5,8 +5,8 @@ import datetime
 # Attempt to import mpi4py.
 # If PYOPTSPARSE_REQUIRE_MPI is set to a recognized positive value, attempt import
 # and raise exception on failure. If set to anything else, no import is attempted.
-if 'PYOPTSPARSE_REQUIRE_MPI' in os.environ:
-    if os.environ['PYOPTSPARSE_REQUIRE_MPI'].lower() in ['always', '1', 'true', 'yes']:
+if "PYOPTSPARSE_REQUIRE_MPI" in os.environ:
+    if os.environ["PYOPTSPARSE_REQUIRE_MPI"].lower() in ["always", "1", "true", "yes"]:
         from paropt import ParOpt as _ParOpt
         from mpi4py import MPI
     else:
@@ -19,12 +19,6 @@ else:
         from mpi4py import MPI
     except ImportError:
         _ParOpt = None
-
-# try:
-#     from paropt import ParOpt as _ParOpt
-#     from mpi4py import MPI
-# except ImportError:
-#     _ParOpt = None
 
 from ..pyOpt_optimizer import Optimizer
 from ..pyOpt_error import Error
--- a/pyoptsparse/pyParOpt/ParOpt.py
+++ b/pyoptsparse/pyParOpt/ParOpt.py
@@ -2,11 +2,29 @@ import numpy as np
 import os
 import datetime
 
-try:
-    from paropt import ParOpt as _ParOpt
-    from mpi4py import MPI
-except ImportError:
-    _ParOpt = None
+# Attempt to import mpi4py.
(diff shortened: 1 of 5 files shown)


pyoptsparse/pyParOpt/ParOpt.py
```diff
@@ -2,11 +2,29 @@
...
+# If PYOPTSPARSE_REQUIRE_MPI is unset, attempt to import mpi4py, but continue on failure
+# with a notification.
+else:
+    try:
+        from paropt import ParOpt as _ParOpt
+        from mpi4py import MPI
+    except ImportError:
+        _ParOpt = None
+
+# try:
```
nwu63, line 23: Please remove the comment block here.

hschilling: Sorry about leaving the comment in. Removed. 

I will also add a test.

I can install black. Not an issue at all.

hschilling: I pushed up a new version with black formatting and also a test

nwu63:
I just took a look at the failed tests.
- For python 2, you could just skip the test since `importlib` does not provide a function called `reload`. As this feature is meant to work with OpenMDAO which only supports py3, this is perfectly fine IMO. Something like a try/except for `importlib.reload`, and put a SkipTest on the except portion.
- For python 3, I forgot that we had removed ParOpt from our test images temporarily. Please skip ParOpt tests for now and I will re-enable them once we start testing ParOpt.

hschilling:
Neil, I could support Python 2. I didn't realize you did still support that:
https://docs.python.org/2/library/functions.html#reload

I am confused about the comment about ParOpt. I see you have tests that use paropt, for example:

https://github.com/mdolab/pyoptsparse/blob/master/test/test_hs015.py#L226

Won't that test fail?

hschilling:
So I could do:

```if sys.version_info[0] == 2:
    reload_func = reload
else:
    reload_func = importlib.reload

class TestNoMPI(unittest.TestCase):

    # Next 3 tests check how the environment variable affects importing MPI
    def test_require_mpi(self):
        os.environ["PYOPTSPARSE_REQUIRE_MPI"] = "1"
        import pyoptsparse.pyOpt_MPI

        reload_func(pyoptsparse.pyOpt_MPI)
        self.assertTrue(inspect.ismodule(pyoptsparse.pyOpt_MPI.MPI))
```

nwu63:
We will drop python 2 support some time this year, but seeing that it's fairly easy to support both, I would suggest we do that with the approach you posted.

We have `SkipTest` in place if we cannot import the optimizer. As ParOpt is not currently installed in our testing images, those tests will simply be skipped. See [here](https://github.com/mdolab/pyoptsparse/blob/master/test/test_hs015.py#L74)

hschilling:
Ah, I see now. I should have looked in that method.

hschilling:
Made the changes based on your comments. Hopefully that does the trick!

nwu63:
`flake8` does more than formatting checks -- it performs syntax and "coding practices" checks. The two warnings we had were
- the ParOpt object imported was unused. This is fine since we are using the import just to determine whether we want to skip the test.
- for Python 3, the py2 `reload` function was undefined. This is also fine and something we have to live with if we are to support both py2 and py3.

I've pushed up a new commit which added inline ignore for these specific flake8 warnings. flake8 should pass now

PR #123, opened 2020-06-09
Update installation instructions for skipping NSGA2

nwu63:
## Purpose
I updated the installation instructions, specifically for when SWIG is unavailable and the user does not require NSGA2.

## Type of change
What types of change is it?

- Documentation update


sseraj:
There would be an import error I think

nwu63:
Correct, because NSGA was not built, the import statement in `__init__.py` will fail. That's the motivation behind this PR.

PR #129, opened 2020-06-18
Cast funcs to real

### LINKED ISSUE #76: Make stored funcs values real instead of complex
## Type of issue
What types of issue is it?

## Description
Due to complex casting for complex step derivatives, the `funcs` values are complex. They should be cast back to real to be saved in the history file.

### Current behavior
The `funcs` values are complex with zero imaginary components

### Expected behavior
They should be real valued

sseraj:
## Purpose

Resolves #76

I added a loop over funcs in _masterFunc2 to cast all values to real. I also removed some now redundant casting that would have occured after the call to _masterFunc2. This includes reverting the commits in #102.

## Type of change

## Testing
Run `test_hs015.py`, which includes new tests in `check_hist_file`.

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimization.py
+++ b/pyoptsparse/pyOpt_optimization.py
@@ -1659,10 +1659,7 @@ class Optimization(object):
         fmt = "    {0:>7d}  {1:{width}s}   {2:>14.6E}   {3:>14.6E}\n"
         for idx, name in enumerate(self.objectives):
             obj = self.objectives[name]
-            value = obj.value
-            if np.iscomplexobj(value):
-                value = value.real
-            text += fmt.format(idx, obj.name, value, obj.optimum, width=num_c)
+            text += fmt.format(idx, obj.name, obj.value, obj.optimum, width=num_c)
 
         # Find the longest name in the variables
         num_c = 0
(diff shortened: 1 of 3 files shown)


nwu63:
Do you know where the complex numbers are being introduced? Instead of patching it like this, I would prefer to fix it at the source. I suspect it's due to some incorrect type passed to the `Gradient` class or something like that.

sseraj:
Actually for `test_hs015.py`, which I was using as a test case, the complex numbers are introduced in `objfunc` from `conval = np.zeros(2, "D")`. I replaced this with `conval = np.zeros(2, "d")`, and there were no complex numbers even without the patch.

nwu63:
The problem is a bit more complex than that.

1. Please keep the test you wrote, but remove the fix where you cast everything to real. If changing the type from `D` to `d` fixes it for `hs015`, let's keep that.

2. The issue is more specifically for when we use complex step to compute derivatives. See for example `test_tp109`. The `funcs` you get from the SNOPT run will be complex. That needs to be addressed and I think it's probably in the complex step part of the code.

sseraj:
Thanks! I was just looking for an example that uses complex step. This patch does not work for `test_tp109` anyway, so I'm going to close this PR and address the root of the problem as you suggested.

PR #130, opened 2020-06-18
Add options and informs tables

nwu63:
## Purpose
I created a custom Sphinx directive, which will generate the options and informs tables for each optimizer, reading from the `defOpts` and `informs` class attributes. In order to make this work, I had to make the following changes:
- added `defOpts` and `informs` as class attributes
- added `raiseError` option in class instantiation, so that when we generate the table we do not need the compiled optimizer modules

For future reference, here are some resources I found helpful:
- [the reference implementation](https://github.com/BU-NU-CLOUD-SP16/Trusted-Platform-Module-nova/blob/master/api-ref/ext/rest_parameters.py) and the [generated docs](https://docs.openstack.org/api-ref/compute/#restore-soft-deleted-instance-restore-action)
- [docutils docs](https://docutils.readthedocs.io/en/sphinx-docs/howto/rst-directives.html)

## Type of change
What types of change is it?

- Documentation update

## Testing
Table is generated locally

## Checklist


nwu63:
It works! Should be ready for review now.

nwu63:
I don't think it's ready for that yet. The most important link is in the docstring so we don't need to look for it on GitHub. I'm happy to talk to whoever willing to do this in the future.

PR #133, opened 2020-06-29
Update ipopt

nwu63:
## Purpose
I updated the installation instructions to install the latest version of IPOPT with pyOptSparse. There are minor changes to the `setup.py` but most of it is just updating the docs. One key change is that I have defaulted to the MUMPS linear solver, since it can be downloaded and installed without any registration required.

Note that the tests may fail since the Docker image has not been updated, but once that is done everything should work just fine. Reviewers are encouraged to follow the instructions and install IPOPT to verify this PR.

## Type of change
What types of change is it?

- Documentation update

## Testing
The IPOPT tests work with the latest version of IPOPT.

## Checklist

### DIFF
--- a/doc/optimizers/pyipopt.rst
+++ b/doc/optimizers/pyipopt.rst
@@ -47,6 +47,7 @@ Here we explain a basic setup using MUMPS as the linear solver, together with ME
    .. code-block:: bash
 
       # build IPOPT
+      cd $IPOPT_DIR
       mkdir build
       cd build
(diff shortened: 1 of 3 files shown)


doc/optimizers/pyipopt.rst
```diff
@@ -1,56 +1,79 @@
...
 
-#. Go to::
+#. Build IPOPT
 
-     Ipopt/ThirdParty/Lapack/
+   .. code-block:: bash
 
-   and run::
+      # build IPOPT
+      mkdir build
```
sseraj, line 50: Add `cd $IPOPT_DIR` for completeness?

nwu63: Done

PR #134, opened 2020-07-01
Add entry point

nwu63:
## Purpose
I updated `setup.py` to have OptView as an `entry_point` which means it will be automatically available on the `$PATH`. There is no longer a need to define an alias for it.

The changes are:
- moved `postprocessing` folder under `pyoptsparse` so that it can be installed and used.
- updated multiple `setup.py` and made `postprocessing` a sub-module
- renamed `__main__` to `main()` so that the entry_point could call the main function for OptView

@johnjasa could you take a look just to make sure I didn't miss anything major here?

## Type of change
What types of change is it?

- Maintenance update

## Testing
Install the package, and check that `optview` should be already available without defining a custom alias. OptView should work as before.

## Checklist


nwu63:
So, it's also possible to not move postprocessing to inside the repo, and instead use the `script` argument in setuptools. We could go that way if needed -- I'm not convinced this is the best approach.

johnjasa:
I checked it out in detail and looks good to me! I think the entry-point use for OptView is much cleaner.

PR #140, opened 2020-07-20
Some fixes for linear constraints in the history file

nwu63:
## Purpose
This PR fixes an issue with linear constraints being stored in the `conInfo` field of the history file. Since `conNames` uses the information in `conInfo`, the linear constraint name also appears there. Lastly, by default `getValues` returns all values, and will attempt to extract the value of this linear constraint, which does not exist in `funcs` since it's linear.

I updated `getConNames` so that it skips linear constraints. I also updated the part where `conInfo` is written to the history file during optimization, so that linear constraints are no longer being written. Finally, a test is added to check for this.

## Type of change
What types of change is it?

## Testing
I updated tp109 to check that the linear constraint is not stored in either `conInfo` or `conNames`, and that `getValues` no longer throws an error.

## Checklist


pyoptsparse/pyOpt_history.py
```diff
@@ -193,10 +193,13 @@ def _processDB(self):
...
+        # metadata
+        self.metadata = self.read("metadata")
+        self.optProb = self.read("optProb")
         # load names
-        self.DVNames = set(self.DVInfo.keys())
-        self.conNames = set(self.conInfo.keys())
-        self.objNames = set(self.objInfo.keys())
+        self.DVNames = set(self.getDVNames())
+        self.conNames = set(self.getConNames())
+        self.objNames = set(self.getObjNames())
```
marcomangano, line 203: Is this also related to py3 update? why were these `.keys()` used when the other functions were available? Anyway, good you updated this

nwu63: The old implementation used the `.keys()` call, but I realized that I wrote these functions. Previously, these functions did the exact same thing, but I updated `getConNames` to skip linear constraints, and here this should be consistent also, so I switched to using these functions.

nwu63:
> It looks good to go, I am just missing one thing: where are we storing the linear constraints info now? You removed the "extras" from both `conNames` and the history file, not sure about what is left.

The linear constraints are not stored explicitly anywhere, since they are not evaluated in the `obj` or `sens`. However, they are accessible via `optProb.constraints.jac` for linear constraints. Given that they are just mat-vec products, you can just do `A*x` to get the constraint value.

marcomangano:
> The linear constraints are not stored explicitly anywhere, since they are not evaluated in the `obj` or `sens`. However, they are accessible via `optProb.constraints.jac` for linear constraints. Given that they are just mat-vec products, you can just do `A*x` to get the constraint value.

Sounds good, maybe we can add this one liner somewhere in the documentation, if it is not there already and you feel it is worth it.

nwu63:
Agreed, the documentation should explain a bit more on how linear constraints are handled and so forth. I will open an issue for this.

PR #144, opened 2020-07-25
Additional linear constraint fixes

nwu63:
## Purpose
This PR fixes two more bugs with the linear constraints:
- casting the constraint dict_keys into a list, so that the dictionary does not mutate during the loop if an item is removed. This bug was not caught by the test since the linear constraint was always the last constraint in the way the test was set up, and therefore was unaffected
- Simplified the way `conNames` was computed, somehow the previous implementation did not work as expected

I also added a test to `getValues` that tests the scale option.

## Type of change
What types of change is it?

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_history.py
+++ b/pyoptsparse/pyOpt_history.py
@@ -436,6 +436,8 @@ class History(object):
             if not self.optProb.constraints[con].linear:
                 conDict[con] = d["funcs"][con]
             else:
+                # the linear constraints are removed from optProb so that scaling works
+                # without needing the linear constraints to be present
                 self.optProb.constraints.pop(con)
         objDict = {}
         for obj in self.objNames:
--- a/pyoptsparse/pyOpt_history.py
+++ b/pyoptsparse/pyOpt_history.py
@@ -257,10 +257,8 @@ class History(object):
         # only do this if we open the file with 'r' flag
         if self.flag != "r":
             return
-        conNames = list(self.conInfo.keys())
-        for con in conNames:
-            if self.optProb.constraints[con].linear:
-                conNames.remove(con)
+        # we remove linear constraints
+        conNames = [con for con in self.conInfo.keys() if not self.optProb.constraints[con].linear]
         return copy.deepcopy(conNames)
 
     def getObjNames(self):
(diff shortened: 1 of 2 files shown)


pyoptsparse/pyOpt_history.py
```diff
@@ -257,10 +257,8 @@ def getConNames(self):
         # only do this if we open the file with 'r' flag
         if self.flag != "r":
             return
-        conNames = list(self.conInfo.keys())
-        for con in conNames:
-            if self.optProb.constraints[con].linear:
-                conNames.remove(con)
+        # we remove linear constraints
+        conNames = [con for con in self.conInfo.keys() if not self.optProb.constraints[con].linear]
```
marcomangano, line 261: This is much cleaner! Just to be sure, this loop generates an identical object to `list()`, right?

pyoptsparse/pyOpt_history.py
```diff
@@ -433,7 +431,7 @@ def _processIterDict(self, d, scale=False):
         These are all "flat" dictionaries, with simple key:value pairs.
         """
         conDict = {}
-        for con in self.optProb.constraints.keys():
+        for con in list(self.optProb.constraints.keys()):
             # linear constraints are not stored in funcs
```
marcomangano, line 435: Now that you explain it the bug is evident, but what kind of issues was it leading to? how did you catch it?

nwu63: You will get an error saying `OrderedDict mutated during iteration`

nwu63:
> Good PR, I am just wondering why we use `optProb.constraints.keys()` in `getObjNames()`. Could we use `conNames` directly or with some complementary list to pop out the linear constraints from `optProb`? Maybe it is just not worth the hassle, just want to hear your take on this

That is a great question. The reason is because, the scaling functions such as `self.optProb._mapContoOpt_Dict` called below actually require all constraints, including linear constraints. So, we just remove those linear constraints from the copy of `optProb` here so that scaling works.

In the future, I will clean this up and use a better implementation.
- No need for `conInfo` etc now that we have `optProb`
- Better implementation to explicitly handle both linear and nonlinear constraints

sseraj:
To clarify, the above discussion is about the function _processIterDict, not getObjNames, is that right?

nwu63:
Yep. `getObjNames` is still using `conInfo` but at some point I will change that too, and rely only on optProb.

Issue #152, opened 2020-08-13
Error installing IPOPT solver


dhill2522:
## Type of issue
What types of issue is it?
_Select the appropriate type(s) that describe this issue_

- Bugfix (non-breaking change which fixes an issue)

## Description

When following the installation instructions for IPOPT I ran into an error when the `IPOPT` folder was at `pyoptsparse/pyIPOPT/IPOPT`. The current instructions worked fine up through step 5. On step 6 I got the following error as it appears that some of the header files are missing.

```bash
In file included from pyoptsparse/pyIPOPT/src/callback.c:36:
pyoptsparse/pyIPOPT/src/hook.h:5:10: fatal error: IpStdCInterface.h: No such file or directory
    5 | #include "IpStdCInterface.h"
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
In file included from pyoptsparse/pyIPOPT/src/pyipoptcoremodule.c:9:
pyoptsparse/pyIPOPT/src/hook.h:5:10: fatal error: IpStdCInterface.h: No such file or directory
    5 | #include "IpStdCInterface.h"
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

### Steps to reproduce issue
1. Follow the IPOPT installation process using `IPOPT_DIR=path/to/pyoptsparse/pyoptsparse/pyIPOPT/IPOPT`. 

### Current behavior
Crashes during step 6.

### Expected behavior
Successful installation

### Code version (if relevant)
**Python version:** 3.8.3

**External dependencies:** …

**Internal packages:** …

dhill2522:
I can put a pull request together to fix this real quick.

PR #153, opened 2020-08-13
fixed the IPOPT installation bug

dhill2522:
## Purpose
Fix an installation issue for the IPOPT solver. This specifically addresses the case when `$IPOPT_DIR` is set to `pyoptsparse/pyoptsparse/pyIPOPT/IPOPT`. The issue is #152 .

## Type of change

### DIFF
--- a/pyoptsparse/pyIPOPT/setup.py
+++ b/pyoptsparse/pyIPOPT/setup.py
@@ -32,7 +32,7 @@ def configuration(parent_package="", top_path=None):
     if os.path.exists("pyoptsparse/pyIPOPT/Ipopt"):
         IPOPT_DIR = os.path.join(top_path, "pyoptsparse/pyIPOPT/Ipopt/")
         IPOPT_LIB = os.path.join(top_path, "pyoptsparse/pyIPOPT/Ipopt/lib")
-        IPOPT_INC = os.path.join(IPOPT_DIR, "include/coin/")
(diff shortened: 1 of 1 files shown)


dhill2522:
If you would rather simply make the change in a different branch rather than in a pull request that is fine with me as well.

nwu63:
Ah yes, thanks for the PR Daniel I see that the include directory is indeed incorrect. The IPOPT source and L44 of setup.py both say that the directory is called `coin-or` and not `coin`. The IPOPT build script we use for testing uses the `IPOPT_DIR` environment variable which is why this was not caught before.

As for the branch, I'm happy to pull this in. However, note that since we squash PR commits, there will be an inconsistency between your master branch and the master branch under `mdolab`, so that if pull your `master` by merging you will likely have an extra commit. We typically submit PR under a dedicated PR branch, which is something you might want to consider in the future. However this does not impact us so I will approve the PR.

dhill2522:
Do you mean the PR description or the issue description? I could not find another PR to compare mine to.

nwu63:
I think he means the PR description, which has been pre-populated with some instructions. Please edit it and fill in the information as necessary. See for example [this PR](https://github.com/mdolab/pyoptsparse/pull/140),

dhill2522:
Ah, thanks. I should have noticed that. 
It is updated now.

PR #156, opened 2020-09-09
Updated black formatting

nwu63:
## Purpose
Black recently released a new (hopefully last) beta version after a year or so, which is why the tests failed on the master branch. Here I updated the code so that it conforms to black again. As far as I can tell, the main change relevant for us is the new capability to handle docstrings, which is what most of the diffs are about.

## Type of change
What types of change is it?

- Formatting

## Checklist

### DIFF
--- a/examples/snopt_bugfix.py
+++ b/examples/snopt_bugfix.py
@@ -23,8 +23,7 @@ def objfunc(xdict):
 
 
 def sens(xdict, funcs):
-    """f(x,y) = (x-3)^2 + xy + (y+4)^2 - 3
-    """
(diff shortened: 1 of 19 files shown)


marcomangano (approved):
I see the changes are very minor and the checks pass, so I believe everything is fine! Might be a good time to start re-formatting other repos?

nwu63:
Yeah, probably a good time if anyone has time. This shows the power of automatic tests, which was able to catch this when we probably would not.

If you format code locally you will have to update black using `pip install black -U`.

sseraj (approved):

PR #159, opened 2020-09-10
Update setup

nwu63:
## Purpose
This PR updates the dependency specification.
- the requirements.txt file has been removed, since the dependencies are specified within setup.py
- the corresponding files for RTD are removed also
- the patch version for dependencies are removed in setup.py

## Type of change
What types of change is it?

- Maintenance

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -42,6 +42,9 @@ extensions = [
 ]
 numpydoc_show_class_members = False
 
+# mock import for autodoc
+autodoc_mock_imports = ["numpy", "scipy", "six", "sqlitedict"]
+
 # tell autodoc to document the __init__ methods
 autoclass_content = "both"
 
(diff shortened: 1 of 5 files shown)


marcomangano (dismissed):

eirikurj (approved):

nwu63:
After several tries, I don't think I can get RTD to work without actually writing out the dependencies in a requirements.txt file, because `setup.py` needs `numpy.distutils` to work. This is the chicken-and-egg problem that prompted the Python community to develop things like `pyproject.toml`, but most projects have yet to support such a thing. Given the confusion over this whole mess, I'm going to delay any work on this and re-submit this PR just to remove the patch versions from dependencies.

In the future, we may want to move to some sort of metadata file such as `setup.cfg` or `pyorject.toml`.

Issue #145, opened 2020-07-26
Update testing framework


nwu63:
## Type of issue
What types of issue is it?
_Select the appropriate type(s) that describe this issue_

- Testing

## Description
We should update the overall testing framework:
- evaluate if we want to add other optimization problems
- share code among tests -- most of them are doing similar things and we should reduce code duplication
- improve overall code coverage

matbesancon:
Not sure if this is related, but testing seems to block when running the command, and seems not to block but with no test being run without failing with the verbose flag, complete error:
```
./test/test_snopt_bugfix.py:TestSNOPTBug.test_opt_bug1 ... None (00:00:0.00, 0 MB)
./test/test_hs015.py:TestHS15.test_ipopt ... None (00:00:0.00, 0 MB)
./test/test_hs015.py:TestHS15.test_nlpqlp ... None (00:00:0.00, 0 MB)
./test/test_hs015.py:TestHS15.test_paropt ... None (00:00:0.00, 0 MB)
./test/test_hs015.py:TestHS15.test_psqp ... None (00:00:0.00, 0 MB)
./test/test_hs015.py:TestHS15.test_slsqp ... None (00:00:0.00, 0 MB)
./test/test_hs015.py:TestHS15.test_snopt ... None (00:00:0.00, 0 MB)
./test/test_snopt_bugfix.py:TestSNOPTBug.test_opt_bug_print_2con ... None (00:00:0.00, 0 MB)
./test/test_hs015.py:TestHS15.test_conmin ... None (00:00:0.00, 0 MB)
./test/test_snopt_bugfix.py:TestSNOPTBug.test_opt ... None (00:00:0.00, 0 MB)
./test/test_require_mpi_env_var.py:TestRequireMPIEnvVar.test_do_not_use_mpi ... FAIL (00:00:0.00, 31 MB)
Traceback (most recent call last):
  File "/home/mbesancon/.local/lib/python3.6/site-packages/testflo/test.py", line 425, in _try_call
    func()
  File "./test/test_require_mpi_env_var.py", line 42, in test_do_not_use_mpi
    import pyoptsparse.pyOpt_MPI
ModuleNotFoundError: No module named 'pyoptsparse'

./test/test_require_mpi_env_var.py:TestRequireMPIEnvVar.test_no_mpi_requirement_given ... FAIL (00:00:0.00, 31 MB)
Traceback (most recent call last):
[...]

PR #162, opened 2020-09-12
Bugfix and testing for unconstrained optimization

### LINKED ISSUE #157: PSQP and CONMIN not working for unconstrained problems
## Type of issue
What types of issue is it?

Bug
## Description
wrappers for CONMIN and PSQP pass incorrectly formatted variables to FORTRAN code for unconstrained problems.
In addition, it appears that both methods also need the "dummy Constraint" as implemented in SLSQP as the code tries to access a constraint variable that doesn't exit (which the introduction of the dummy constraint fixes).

I attached modified code for pyCONMIN.py and pyPSQP.py that fix the problems
### Steps to reproduce issue
1. Run attached test script implementing Rosenbrock's function (modified from pyOpt test script)
2. …
3. …

### Current behavior
Python throws an error message when calling the FORTRAN wrapper. Once that's fixed, non-existent variable "gcon" causes another error (fixed with the introduction of the dummy constraint)

### Expected behavior
no error

### Code version (if relevant)
**Python version:** 3.7

[bugfix.zip](https://github.com/mdolab/pyoptsparse/files/5198120/bugfix.zip)

marcomangano:
## Purpose
As pointed out in #157, few optimizers in pyOptSparse do not natively support an unconstrained optimization problem formulation.
We fix this issue using `dummyConstraint` option from the `Optimization()` class, indeed providing a dummy constraint to the algorithm to complete the input set.
As this kind of optimization problem was untested, the option was not turned on for PSQP, CONMIN, and NLPQLP.

Now the feature is enabled, supporting unconstrained optimization problems throughout pyOptSparse. 
Moreover, I added a new test based on the Rosenbrock function to verify that all our optimizers support unconstrained problems.

Closes #157 and partially addresses #145.

## Type of change

## Testing
Run the entire test suite or just the newly added `test_rosenbrock.py`

## Checklist

### DIFF
--- a/test/test_rosenbrock.py
+++ b/test/test_rosenbrock.py
@@ -103,12 +103,9 @@ class TestRosenbrock(unittest.TestCase):
         sol_xvars = [sol.variables["xvars"][i].value for i in range(n)]
 
         assert_allclose(sol_xvars, dv["xvars"], atol=tol, rtol=tol)
-        # we check either optimum via try/except
-        # try:
+
         assert_allclose(sol.objectives["obj"].value, self.fStar1, atol=tol, rtol=tol)
         assert_allclose(dv["xvars"], self.xStar1, atol=tol, rtol=tol)
-        # except AssertionError:
-        #     pass
 
     def check_hist_file(self, optimizer, tol):
         """
--- a/test/test_rosenbrock.py
+++ b/test/test_rosenbrock.py
@@ -122,20 +122,17 @@ class TestRosenbrock(unittest.TestCase):
 
         # Info checks
         self.assertEqual(hist.getDVNames(), ["xvars"])
-        # self.assertEqual(hist.getConNames(), ["con"])
         self.assertEqual(hist.getObjNames(), ["obj"])
         dvInfo = hist.getDVInfo()
         self.assertEqual(len(dvInfo), 1)
         self.assertEqual(dvInfo["xvars"], hist.getDVInfo(key="xvars"))
         conInfo = hist.getConInfo()
         self.assertEqual(len(conInfo), 0)
-        # self.assertEqual(conInfo["con"], hist.getConInfo(key="con"))
         objInfo = hist.getObjInfo()
         self.assertEqual(len(objInfo), 1)
         self.assertEqual(objInfo["obj"], hist.getObjInfo(key="obj"))
         for key in ["lower", "upper", "scale"]:
             self.assertIn(key, dvInfo["xvars"])
-            # self.assertIn(key, conInfo["con"])
         self.assertIn("scale", objInfo["obj"])
 
         # callCounter checks
--- a/test/test_rosenbrock.py
+++ b/test/test_rosenbrock.py
@@ -1,0 +1,249 @@
+"""Test solution of Rosenbrock problem"""
+
+import unittest
+import numpy as np
+from numpy.testing import assert_allclose
+from pyoptsparse import Optimization, OPT, History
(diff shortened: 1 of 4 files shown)


test/test_rosenbrock.py
```diff
@@ -0,0 +1,249 @@
...
+        # Check Solution
+        self.fStar1 = 0.0
+
+        self.xStar1 = np.ones(n)
+
+        dv = sol.getDVs()
+        sol_xvars = [sol.variables["xvars"][i].value for i in range(n)]
+
+        assert_allclose(sol_xvars, dv["xvars"], atol=tol, rtol=tol)
+        # we check either optimum via try/except
```
nwu63, line 106: Please remove these comments

test/test_rosenbrock.py
```diff
@@ -0,0 +1,249 @@
...
+        # try:
+        assert_allclose(sol.objectives["obj"].value, self.fStar1, atol=tol, rtol=tol)
+        assert_allclose(dv["xvars"], self.xStar1, atol=tol, rtol=tol)
+        # except AssertionError:
```
nwu63, line 110: and of course these

nwu63 (commented):
The rest looks good. I've verified that the test you added fails for the current `master` branch, for CONMIN, PSQP and NLPQLP. Good work

test/test_rosenbrock.py
marcomangano: Yes thanks, I missed them. I copied and adapted `test_hs071.py` for this test. I removed the try/except statement but not the comments

nwu63 (approved):
Yep looks good. But you can see why I wanted to refactor the tests, there is now significant code duplication between the different tests, so we should pull these out into a util package or something like that.

marcomangano:
Agreed, probably we can move the different objective functions under a separate class / method and have a single set of functions that call pyOptSparse from a single file. I haven't checked the other tests, but I believe `hs_071` and `hs_015` have minimal differences, and of course in this new test I changed a few hardcoded values and commented out all the calls that refer to the `"con"` keyword.

test/test_rosenbrock.py
```diff
@@ -0,0 +1,246 @@
...
+        metadata = hist.getMetadata()
+        self.assertEqual(metadata["optimizer"], optimizer)
+        metadata_def_keys = ["optName", "optOptions", "nprocs", "startTime", "endTime", "optTime", "version"]
+        for key in metadata_def_keys:
+            self.assertIn(key, metadata)
+        hist.getOptProb()
+
+        # Info checks
+        self.assertEqual(hist.getDVNames(), ["xvars"])
+        # self.assertEqual(hist.getConNames(), ["con"])
```
sseraj, line 125: Do we want to keep these lines as comments or remove them?

sseraj (commented):
I have also checked that the relevant tests fail without the patch.

test/test_rosenbrock.py
marcomangano: These come from `test_hs_071` as well. I don't have any strong opinion about that , I left them as "placeholders" for a later refactoring, but we can take them off, up to you guys!

sseraj: @nwu63 Any thoughts on this?

nwu63: I think we should remove these

sseraj (approved):

nwu63 (approved):
@marcomangano it's 1am, go sleep!

PR #164, opened 2020-09-22
Remove python 2 support

nwu63:
## Purpose
This PR removes support for Python 2. Once this is merged, pyOptSparse will no longer work with Python 2.
The main change was just removing the `six` package and changing the syntax to use the native py3 iterators.

## Type of change
What types of change is it?

- Breaking change (non-backwards-compatible fix or feature)

## Testing
All existing tests pass on py3.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,15 +55,15 @@ install:
       docker exec -it app /bin/bash -c "cp -r \$HOME/NLPQLP/* $NLPQLP_DIR/ && cp -r \$HOME/SNOPT/* $SNOPT_DIR/";
     fi
   # Build and install
-  - docker exec -it app /bin/bash -c ". \$HOME/.bashrc && cd $DOCKER_WORKING_DIR && pip install -e ."
+  - docker exec -it app /bin/bash -c ". \$HOME/.bashrc_mdolab && cd $DOCKER_WORKING_DIR && pip install -e ."
 script:
   # We need to source the mdolab bashrc before running anything
-  - docker exec -it app /bin/bash -c ". \$HOME/.bashrc && cd $DOCKER_WORKING_DIR && testflo --pre_announce -v . --coverage --coverpkg $REPO_NAME"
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -11,19 +11,19 @@ services:
...
-    - DOCKER_TAG=u18-gcc-ompi-py2-stable
-    - DOCKER_TAG=u18-gcc-ompi-py3-latest
-    - DOCKER_TAG=u18-gcc-ompi-py3-stable
-    - DOCKER_TAG=u20-gcc-ompi-py3-latest
-    - DOCKER_TAG=u20-gcc-ompi-py3-stable
+    # - DOCKER_TAG=c7-gcc-ompi-latest
+    - DOCKER_TAG=c7-intel-impi-latest
+    - DOCKER_TAG=u18-gcc-ompi-stable
+    - DOCKER_TAG=u18-gcc-ompi-latest
+    - DOCKER_TAG=u18-gcc-ompi-stable
```
marcomangano, line 24: Since we removed the py2 test, I feel we have a duplicate `u18-gcc-ompi-stable` build here

.travis.yml
```diff
@@ -11,19 +11,19 @@ services:
   - docker
 env:
   global:
-    - REPO_NAME=pyoptsparse
+    - REPO_NAME=$(eval basename $TRAVIS_REPO_SLUG)
```
marcomangano, line 14: Just a curiosity here, how does this command work? what kind of env variable is `$TRAVIS_REPO_SLUG`?

marcomangano (commented):
Check the comment about the extra tag below, looks good otherwise

.travis.yml
nwu63: Yes very good catch.

.travis.yml
nwu63: It's set to `mdolab/pyoptsparse` for this repo. `basename` picks out the last portion after all the `/` so it's a convenient way to get the repo name programmatically.

marcomangano (approved):

sseraj (approved):

PR #173, opened 2020-10-08
Update examples

nwu63:
## Purpose
This PR updates some of the examples that were deprecated. It also changes the type of error raised when the SqliteDict is not found.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
All the examples work now, especially when using non-default command line arguments.

## Checklist


Xiaosong2105 (commented):
The updated way of constructing funcsSens dictionary looks good.
The pyoptsparse.History and sqlitedict.SqliteDict are basically doing the same thing, correct?

examples/rosenbrock.py
```diff
@@ -45,87 +32,50 @@ def sensfunc(xdict, funcsDict):
...
-        opt.setOption("Major iterations limit", 10)
-        solSnopt1 = opt(optProb, sens=sens, sensMode="pgc", storeHistory="opt_hist")
 
-        # Now we are allowed to do 50
-        opt.setOption("Major iterations limit", 50)
-        if testHist == "hot":
-            solSnopt2 = opt(optProb, sens=sens, sensMode=sensMode, hotStart="opt_hist", storeHistory="opt_hist")
-        else:
-            solSnopt2 = opt(optProb, sens=sens, sensMode=sensMode, coldStart="opt_hist", storeHistory="opt_hist")
+if args.testHotstart:
```
marcomangano, line 50: Is this the best practice to map a specific option to different optimizers? What if we are dealing with multiple options and not just `max_iter`?

marcomangano (commented):
See below, looks good for the rest.

nwu63:
> The pyoptsparse.History and sqlitedict.SqliteDict are basically doing the same thing, correct?
Almost. The `History` class is basically an instance of the SqliteDict, but with its specific format.

examples/rosenbrock.py
nwu63: Yeah this is a good point. Maybe we should restrict this to run with SLSQP only, and then everything would be simpler.

Xiaosong2105 (approved):

examples/rosenbrock.py
nwu63: I've updated the rosenbrock example to only run SLSQP.

marcomangano (approved):

PR #175, opened 2020-10-18
Update docs

nwu63:
## Purpose
I removed the old theme and updated `conf.py` to use the new configuration.

## Type of change
What types of change is it?

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
N/A

## Checklist


doc/conf.py
```diff
@@ -1,8 +1,4 @@
-# Configuration file for the Sphinx documentation builder.
-#
-# This file only contains a selection of the most common options. For a full
-# list see the documentation:
-# https://www.sphinx-doc.org/en/master/usage/configuration.html
+from sphinx_mdolab_theme.config import *
```
sseraj, line 1: I'm surprised flake8 isn't complaining about this. I do think this is an appropriate situation to use a star import though.

sseraj (commented):
Is the second [`sys.path.insert`](https://github.com/mdolab/pyoptsparse/blob/5b90fff0aee7b02fcf692cc9f1014d4552a897ee/doc/conf.py#L13) still necessary?

doc/conf.py
nwu63: The reason it did not complain is because I have manually excluded it: https://github.com/mdolab/pyoptsparse/blob/master/.flake8#L11

Right now it is using the custom `.flake8` file, maybe I should add these settings to the shared repo?

nwu63:
> Is the second [`sys.path.insert`](https://github.com/mdolab/pyoptsparse/blob/5b90fff0aee7b02fcf692cc9f1014d4552a897ee/doc/conf.py#L13) still necessary?

Yes, because there is the `optimizertable` extension which is included in this repo. Eventually I hope to also make that part of the sphinx repo so that it's available to other repos as well.

doc/conf.py
sseraj: I think it makes sense to add it to the shared repo.

sseraj (approved):

marcomangano (approved):

PR #184, opened 2020-11-08
Include non-objective/constraint functions in history output

A-Gray-94:
## Purpose
Currently, the `getValues` method in the `pyOpt_history` class only returns the histories of functions that are used as objectives or constraints. However, sometimes there are other functions of interest not used as either that user may want to track the history of anyway. This fix allows this by adding all the functions in the `'funcs'` dict to the list of those that can be returned by `getValues`.

You may want to make some changes to this, I'm not sure if including everything from `self.db["0"]["funcs"]` automatically includes all constraints and objectives or not. Are there also tests that would be affected by this change?

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_history.py
+++ b/pyoptsparse/pyOpt_history.py
@@ -532,7 +532,8 @@ class History(object):
             return
 
         allNames = (
-            self.DVNames.union(self.conNames)
+            self.DVNames.union(self.db["0"]["funcs"].keys())
+            .union(self.conNames)
             .union(self.objNames)
             .union(self.iterKeys)
             .difference(set(["funcs", "funcsSens", "xuser"]))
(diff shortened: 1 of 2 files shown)


nwu63:
I just pushed up some commits. There is now an attribute called `extraFuncsNames` which stores the set of "extra" funcs names. There is an accompanying function to retrieve this list called `getExtraFuncsNames()`, and the values can be retrieved using the `getValues` function as expected.

A-Gray-94:
Great, thanks! I'm happy with this if you are

nwu63 (approved):
Would prefer another approval before we merge, since I worked on this PR.

marcomangano (approved):
Looks sweet now! The approach with `.difference` makes a lot of sense, thanks for adding the test! Not sure why the coverage has reduced, looks good to go for me

nwu63:
The coverage was reduced because this is a public PR, meaning some tests that need SNOPT etc are not run. Not sure how to deal with that..

PR #185, opened 2020-11-12
Update for SNOPT v7.7.5

nwu63:
## Purpose
This PR extends support for SNOPT v7.7.5, which has the `snLog` call signatures restored so there was no need to modify the f2py interface. Changes are:
- Added `Backoff factor` option to control the backtracking factor during line search when encountering failed function evaluations
- Added `Proximal iterations limit` option for controlling the number of iterations in solving the proximal point problem
- The default value for this option has been changed from 200 to 10000 to ensure we solve the problem to optimality. This was also the default behaviour in the older SNOPT v7.2

Note that since the `Backoff factor` did not exist in v7.7.1, attempting to set it via `optOptions` will result in SNOPT saying 
```
XXX  Keyword not recognized:         BACKOFF
```
however, this is simply a warning and SNOPT will carry on as usual.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
All tests run, both with v7.7.1 and v7.7.5.

## Checklist

### DIFF
--- a/doc/optimizers/pysnopt.rst
+++ b/doc/optimizers/pysnopt.rst
@@ -16,8 +16,9 @@ Installation
 SNOPT is available for purchase `here 
 <http://www.sbsi-sol-optimize.com/asp/sol_snopt.htm>`_. Upon purchase, you should receive a zip file. Within the zip file, there is a folder called ``src``. To use SNOPT with pyoptsparse, paste all files from ``src`` except snopth.f into ``pyoptsparse/pySNOPT/source``.
 
-From v2.0 onwards, only SNOPT v7.7 is officially supported.
+From v2.0 onwards, only SNOPT v7.7.x is officially supported.
 To use pyOptSparse with previous versions of SNOPT, please checkout release v1.2.
+The current version of SNOPT being tested is v7.7.1, although v7.7.5 is also expected to work.
 
 
 Options
--- a/doc/optimizers/pysnopt.rst
+++ b/doc/optimizers/pysnopt.rst
@@ -18,7 +18,7 @@ SNOPT is available for purchase `here
 
 From v2.0 onwards, only SNOPT v7.7.x is officially supported.
 To use pyOptSparse with previous versions of SNOPT, please checkout release v1.2.
-The current version of SNOPT being tested is v7.7.1, although v7.7.5 is also expected to work.
+The current version of SNOPT being tested is v7.7.5, although v7.7.1 is also expected to work.
 
 
 Options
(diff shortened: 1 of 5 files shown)


nwu63:
Note that due to recent changes with Travis, builds are no longer able to run. Until we sort out this issue (likely by migrating to an alternate service), please verify the PR locally (perhaps by using the `gh` CLI program to pull down this PR).

marcomangano (approved):
I confirm that all the tests are passing locally. Looking forward to play with these newly available options! I bet we can get some more accuracy and efficiency for some difficult cases.

marcomangano:
Btw, where is the SNOPT warning coming from @nwu63 ? the `f2py` layer? Not a priority but we could fix that later on

nwu63:
> Btw, where is the SNOPT warning coming from @nwu63 ? the `f2py` layer? Not a priority but we could fix that later on

No no this is a warning printed in the `SNOPT_print.out` file at the top, where it lists all the modified options.

doc/optimizers/pysnopt.rst
```diff
@@ -16,8 +16,9 @@ Installation
 SNOPT is available for purchase `here 
 <http://www.sbsi-sol-optimize.com/asp/sol_snopt.htm>`_. Upon purchase, you should receive a zip file. Within the zip file, there is a folder called ``src``. To use SNOPT with pyoptsparse, paste all files from ``src`` except snopth.f into ``pyoptsparse/pySNOPT/source``.
 
-From v2.0 onwards, only SNOPT v7.7 is officially supported.
+From v2.0 onwards, only SNOPT v7.7.x is officially supported.
 To use pyOptSparse with previous versions of SNOPT, please checkout release v1.2.
+The current version of SNOPT being tested is v7.7.1, although v7.7.5 is also expected to work.
```
sseraj, line 21: Can you explain what you mean by this? I saw that  you updated the Docker image to use 7.7.5.
Should we check this PR with both versions?

nwu63: Oh yeah good point, I wrote this before deciding that we should get Docker to build with 7.7.5. From a CI point of view, I don't want to test multiple SNOPT versions, so I guess we've settled on 7.7.5 for that. For manually testing this PR, yes you should test with both versions and that's what I did.

I will update the documentation to say we test 7.7.5.

marcomangano (approved):

sseraj (approved):

PR #188, opened 2020-11-22
Added option to hide default values in docs

### LINKED ISSUE #135: Check default optimizer options
## Type of issue
What types of issue is it?

## Description
The default options stored in the Python wrapper is probably out of date for those optimizers that are still being developed. Check against newest versions and update as necessary.

Also, the behaviour is different across optimizers
- for some optimizers, the default options are used (e.g. IPOPT)
- for others, the default options are not used (e.g. SNOPT, ParOpt)

We need a clear policy on what exactly we are doing, and document this. I think we should NOT store/set defaults for optimizers where we do not provide source code, since they can change defaults at any time and cause inconsistencies with the wrapper. For those optimizers where we provide the source code, we should have the default options accessible in Python, and documented (via options table).

Lastly, there may be default options where we override them in pyOptSparse (think certain options in SNOPT for example). These need to be documented and explained.

nwu63:
## Purpose
I updated the documentation to explain how SNOPT default options come from the optimizer and are not set by pySNOPT.
I have hidden the default values for SNOPT by modifying the Sphinx extension.
I also checked that other optimizers are using the default values, and that the documentation is consistent with the code.
Closes #135.

## Type of change
What types of change is it?

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

### DIFF
--- a/doc/ext/optimizertable.py
+++ b/doc/ext/optimizertable.py
@@ -21,6 +21,7 @@ class OptimizerTable(Table):
         "type": directives.uri,
         "header": str,
         "widths": directives.positive_int_list,
+        "nodefvalues": directives.flag,
     }
 
     def get_options_informs(self, cls):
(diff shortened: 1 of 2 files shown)


Xiaosong2105 (approved):

marcomangano (approved):
Very good!! Thanks for taking care of this, this way is much more rigorous

sseraj (approved):

PR #193, opened 2021-01-08
Switch to using BaseSolver

### LINKED ISSUE #189: Cleaner handling of SNOPT options
# Description
SNOPT has three sets of options
- Options from Fortran with default values
- Options from Fortran with default values overridden in Python 
	- `Proximal iterations limit`
	- `Print file` and `Summary file`, which do not have default values in Fortran and must be set. Here the default values in Python are used unless specified by the user
- Options from Python that are not part of SNOPT
	- `Save major iteration variables` which exists in Python only

These should be described in more detail. We probably want separate dictionaries and special handling to make the code more clear what is happening.

### LINKED ISSUE #190: Consider removing default IPOPT options in Python
# Description
Currently, the IPOPT default options in Python are always set and used. This is potentially problematic as the list could go out of sync with the ones in IPOPT, and this causes extra effort for us to maintain this list. If we support multiple versions of IPOPT and they changed options in between, this would cause additional headaches.

We should consider adopting the approach in SNOPT, where we do not set any default options and let the optimizer take care of it. Any options set by the user should also get passed to the optimizer, but we do not independently maintain a list of default values.

nwu63:
## Purpose
This is a fairly substantial PR that does the following:
- Switch to using baseclasses, primarily for options handling. This was finally doable now that baseclasses is published on PyPI
- Closes #190 by removing default options for IPOPT (except for a few essential ones). Did the same for SNOPT also
- Also closes #189 by clearly defining `self.specialOptions` and `self.pythonOptions`.

Note that this PR depends on mdolab/baseclasses#31. This is a fairly large PR so please take your time to review carefully. Once this is merged, I plan to also update the options table generation to use the static method approach, which will be a lot cleaner than what is done currently.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
No new tests but existing tests pass.

## Checklist


pyoptsparse/pyALPSO/pyALPSO.py
```diff
@@ -226,7 +226,7 @@ def _on_setOption(self, name, value):
             else:
                 raise ValueError("parallel_type must be either '' or 'EXT'.")
 
-    def _on_getOption(self, name, value):
```
marcomangano, line 229: We did not caught this before because the `_on_getOption` methods are mostly used (except for SNOPT, IPOPT, ...) to overwrite the same method from parent `Optimimzer` class right?
Is there a more elegant way to check for the existence of this method in the single optimizer wrappers? Does using a try/except approach become more cumbersome?

pyoptsparse/pyALPSO/pyALPSO.py
```diff
@@ -95,7 +95,9 @@ def __init__(self, *args, **kwargs):
             "parallelType": [str, ""],  # Type of parallelization ('' or 'EXT')
         }
         self.informs = {}
-        Optimizer.__init__(self, "ALPSO", category, self.defOpts, self.informs, *args, **kwargs)
+        super().__init__(
+            self, "ALPSO", category=category, defOptions=self.defOpts, informs=self.informs, options=options
```
marcomangano: Random: are we actually using the `category` attribute anywhere?

pyoptsparse/pyOpt_optimizer.py
```diff
@@ -798,9 +805,9 @@ def setOption(self, name, value=None):
             Variable value to set.
         """
 
-        # Call the optimizer specific routine
+        super().setOption(name, value)
```
marcomangano, line 808: I was wondering why this was previously set after `_on_setOption`
To clarify, `super().setOption()` sets the options in the python layer via the inherited `BaseSolver` method, while `_on_setOption` is the actual optimizer routine that (when available, eg. for SNOPT and IPOPT) passes the options from the python layer to the optimizer "binary", right?

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -676,30 +590,20 @@ def _set_snopt_options(self, iPrint, iSumm, cw, iw, rw):
...
-                        snopt.snset("Warm start", iPrint, iSumm, inform, cw, iw, rw)
-                elif name == "problem type":
-                    if value == "Minimize":
-                        snopt.snset("Minimize", iPrint, iSumm, inform, cw, iw, rw)
-                    elif value == "Maximize":
-                        snopt.snset("Maximize", iPrint, iSumm, inform, cw, iw, rw)
-                    elif value == "Feasible point":
-                        snopt.snset("Feasible point", iPrint, iSumm, inform, cw, iw, rw)
-                elif name == "print file":
+                if name == "Problem Type":
```
marcomangano, line 598: I don't really get what is going on here, can you elaborate a bit? You are removing some calls to inner `snopt` functions so I want to double check

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -676,30 +590,20 @@ def _set_snopt_options(self, iPrint, iSumm, cw, iw, rw):
...
         """
-
         # Set Options from the local options dictionary
         # ---------------------------------------------
         inform = np.array([-1], np.intc)
-        for name, value in self.set_options.items():
-            if name == "iprint" or name == "isumm":
+        for name, value in self.options.items():
+            # these do not get set using snset
+            if name in self.specialOptions or name in self.pythonOptions:
```
marcomangano, line 594: Are you sure you are not skipping stuff here? it seems like you skip the `specialOptions` with `continue` but then you are trying to use some of them (`Print file`) to send stuff to SNOPT via `snset`

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -676,30 +590,20 @@ def _set_snopt_options(self, iPrint, iSumm, cw, iw, rw):
...
-                    if value == "Minimize":
-                        snopt.snset("Minimize", iPrint, iSumm, inform, cw, iw, rw)
-                    elif value == "Maximize":
-                        snopt.snset("Maximize", iPrint, iSumm, inform, cw, iw, rw)
-                    elif value == "Feasible point":
-                        snopt.snset("Feasible point", iPrint, iSumm, inform, cw, iw, rw)
-                elif name == "print file":
+                if name == "Problem Type":
+                    snopt.snset(value, iPrint, iSumm, inform, cw, iw, rw)
+                elif name == "Print file":
```
marcomangano, line 600: Double checking again: the `Print file`,  `Summary file`,  .. are not passed directly to the fortran layer as options, but they are passed directly to SNOPT subroutines right? The outfiles are set up in the `__call__` method above if I get it right

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -612,6 +612,20 @@ def _set_snopt_options(self, iPrint, iSumm, cw, iw, rw):
 
         return
 
+    def _on_setOption(self, name, value):
```
marcomangano: Yeah as I commented before I suppose these need to be here as placeholders to not trigger the error in the parent class right? Is this really best practice?

marcomangano (commented):
This looks like a necessary step up, good! Check my comments for clarifications before we move ahead

pyoptsparse/pyALPSO/pyALPSO.py
nwu63: This was always a bug, but was never caught because previously the `getOption` function defined in the Optimizer class would first `return` then execute `_on_getOption`, so obviously the optimizer-specific `_on_getOption` was never executed. I have swapped the order so that the return statement is at the end of the function.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: This is not necessarily the best practice, but is done more common in other more traditional OOP languages such as C/C++, and is what's usually referred to as an "abstract class" where you define the functions that should be implemented, interfaces, class fields etc without actually implementing anything, leaving all to the inherited classes. This is not exactly Pythonic, we can change it so that the default implementation in Optimizer would simply "pass" everything, and any class that needs more can implement their own. That way, we do not need each inherited class to define the function unless they need something extra, and the base implementation would not throw errors. I will do this.

pyoptsparse/pyALPSO/pyALPSO.py
nwu63: Nope. This was designed to be kind of an attribute that gets queried I think, but it is not currently used. Nobody will instantiate "SNOPT", then query "SNOPT.category" and be surprised to find out that it is a local optimizer... but I'm not inclined to remove it since it might be useful. Let me know if you think otherwise.

nwu63: Plus, `baseclasses` now require a category defined so this fits nicely with that.

pyoptsparse/pyOpt_optimizer.py
nwu63: Not exactly. `super().setOption()` calls the function in BaseSolver, which just sets the options dictionary. Several optimizers maintained a separate `set_options` variable which stores the options which are set by the user rather than by default. These are then passed to the optimizers at a later place (usually during `__call__` since it requires an optimizer instance).

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: Yes I made a mistake here. Those two should not be in `specialOptions`, but also I had to adjust the capitalization for it to work with SNOPT. I ended up adding a `caseSensitiveOptions` param to BaseSolver to force all options to be case sensitive since some optimizers expect it to be, and this was the previous behaviour of the Optimizer class.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: I have removed the `Start` call here, since they were duplicate. The `Start` option can be set in one of two ways, either via the `Cold start` etc. option, or when calling `snkerc` (this one is mandatory and has always been done). I just removed this extra one here since it's not actually necessary.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: Yep the openunit/closeunit actually handles the file name. We just pass the file number (`iSumm` and `iPrint`) to SNOPT, and that's what this section does.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: The `Problem Type` option has just been simplified, since we were setting the same string as before I just got rid of the if statements which were unnecessary. The code is functionally the same.

pyoptsparse/pyALPSO/pyALPSO.py
marcomangano: Yeah, I was wandering if we were actually using that attribute in `baseclasses` as well. I agree we should not remove this.

pyoptsparse/pyOpt_optimizer.py
marcomangano: Ok yeah I got the inherited method and I now I follow how the info is passed, I did not realize that the child `pySNOPT` instance overwrites `_on_setOption` in the inherited `setOption()`, sometimes I miss these OOP tricks

pyoptsparse/pySNOPT/pySNOPT.py
marcomangano: Cool, defaulting to lower case and using the `caseSensitiveOptions` flag only when needed makes sense to me, still support the choice you made for `baseclasses`

pyoptsparse/pyALPSO/pyALPSO.py
marcomangano: Oh gotcha, so basically we are just picking from the dict we set in `setOption()`. I just checked and the `_on_getOption` is a `pass` function in every optimizer wrapper we support now, which further explains why we never bothered to catch this. I will look at the refactoring you made but I feel it makes much more sense to just have everything passing in the parent `Optimizer` class and in case add more, rather than having to override the error in every wrapper

pyoptsparse/pySNOPT/pySNOPT.py
marcomangano: I see the doubled `start` now, makes sense to just keep the `snkerc` one. I also agree with the `Problem Type` simplification, I am just missing where that option is defined now, I am probably messing between commits, but could you point me at that?

pyoptsparse/pySNOPT/pySNOPT.py
marcomangano: `iSumm` and `iPrint` are handled by Fortran at every optimizer iteration? I'll check the manual in case

pyoptsparse/pySNOPT/pySNOPT.py
marcomangano: Ahaha nevermind, I see you added it back in one of the last commits, all good here!

pyoptsparse/pySNOPT/pySNOPT.py
marcomangano: I confirm that  I am onboard with [this commit](https://github.com/mdolab/pyoptsparse/pull/193/commits/868f71259eac4103153b358c73bb52eb9689d5d3), noice!

pyoptsparse/pyOpt_optimizer.py
nwu63: Yeah it is overloaded in the child classes. Very few optimizers use those but they are there.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: Not at every iteration, but used for setting options. If you check `pySNOPT` you can see where `iSumm` and `iPrint` are passed to SNOPT.

pyoptsparse/pyALPSO/pyALPSO.py
nwu63: Yep agreed, I've updated this now.

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -664,27 +585,17 @@ def _set_snopt_options(self, iPrint, iSumm, cw, iw, rw):
...
         inform = np.array([-1], np.intc)
-        for name, value in self.set_options:
-            if name == "iPrint" or name == "iSumm":
+        for name, value in self.options.items():
+            # these do not get set using snset
+            if name in self.specialOptions or name in self.pythonOptions:
                 continue
 
             if isinstance(value, str):
-                if name == "Start":
```
sseraj, line 676: Don't you still need this part to set ``"Start"``?

sseraj (commented):
This looks good. A few comments:

pyoptsparse/pyALPSO/pyALPSO.py
```diff
@@ -223,10 +220,10 @@ def _on_setOption(self, name, value):
                 except ImportError:
                     raise ImportError("pyALPSO: ALPSO EXT shared library failed to import.")
 
-            else:
-                raise ValueError("parallel_type must be either '' or 'EXT'.")
+            elif value != "":
+                raise ValueError("parallelType must be either '' or 'EXT'.")
```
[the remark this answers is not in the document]
nwu63: Agreed, will fix.

pyoptsparse/pyALPSO/pyALPSO.py
```diff
@@ -223,10 +220,10 @@ def _on_setOption(self, name, value):
...
 
-    def _on_getOption(self, name, value):
+    def _on_getOption(self, name):
```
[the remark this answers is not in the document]
nwu63: Whoops will delete

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: Nope, `Start` can be set in two places, either as an argument to `snkerc` (this is mandatory) or a separate option `Cold start` or `Warm start` etc passed to `snset` (this will overwrite the `snkerc` option). Previously, the value for `Start` would be used to set both which is redundant, I've removed that now.

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -390,6 +310,7 @@ def __call__(
                 ncon = 1
             else:
                 ncon = len(indices)
+            print(self.options)
```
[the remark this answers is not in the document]
nwu63: Whoops I should've removed that.

pyoptsparse/pyOpt_optimizer.py
```diff
@@ -852,19 +840,16 @@ def getOption(self, name):
...
         """
 
-        if name in self.options["defaults"]:
-            return self.options[name][1]
-        else:
-            raise Error("Received an unknown option: %s." % repr(name))
-
-        # Now call the optimizer specific routine
+        # Call the optimizer specific routine
         self._on_getOption(name)
```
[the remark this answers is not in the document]
nwu63: Inherited from pyOpt, it's like "commands to execute upon setting optimizer options" type of thing, so I think the name is okay.

sseraj (approved):

marcomangano (approved):
We should coordinate the merge of this PR with [the `baseclasses` one](https://github.com/mdolab/baseclasses/pull/31) today so we avoid buildbot failures, the `install_requires` has been already updated.

nwu63:
Agreed, once baseclasses is merged I will manually trigger a test, and if all passes I would be okay with merging this as long as reviewers are happy.

marcomangano:
I guess we can go for the merge + release?

nwu63:
Yep I think this is ready. Once merged I will make a new release.

PR #194, opened 2021-01-11
Added references

nwu63:
## Purpose
This PR adds a bib file containing publications that have used pyOptSparse in the past. In order to render these, `sphinxcontrib-bibtex` has been added as a dependency for building the docs locally.

I also updated the README slightly to highlight the fact that it integrates into several other frameworks.

## Type of change
What types of change is it?

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
N/A

## Checklist


doc/citation.rst
```diff
@@ -22,3 +22,53 @@ To cite this paper, you can use the following BibTeX entry:
         title = {pyOptSparse: A Python framework for large-scale constrained nonlinear optimization of sparse systems},
         journal = {Journal of Open Source Software}
     }
+
+
+Applications
+------------
+pyOptSparse has been used extensively in the field of engineering design optimization.
+The follow is a non-exhaustive list of works that have used pyOptSparse.
```
marcomangano, line 30: the following?

doc/pyoptsparse.bib
```diff
@@ -0,0 +1,558 @@
...
+@article{Madsen2019a,
+    author  = {Mads H. Aa. Madsen and Frederik Zahle and Niels N. S{\o}rensen and Joaquim R. R. A. Martins},
+    doi     = {10.5194/wes-4-163-2019},
+    journal = {Wind Energy Science},
+    month   = {April},
+    pages   = {163--192},
+    title   = {Multipoint high-fidelity {CFD}-based aerodynamic shape optimization of a 10 {MW} wind turbine},
+    volume  = {4},
+    year    = {2019}
+}
```
marcomangano, line 403: I suppose this paper kinda used MACH-Aero, although for a novel application

marcomangano (commented):
Wow, when you put everything together it actually starts to be a long list!
See my few comments below.
In my opinion this could be in a separate `rst` file from the Citation (right before maybe?) but it is up to you.

nwu63:
> In my opinion this could be in a separate `rst` file from the Citation (right before maybe?) but it is up to you.

Good point, any idea what to call this page? `Past Applications`?

doc/pyoptsparse.bib
nwu63: Right forgot about that, thanks

marcomangano (approved):

doc/conf.py
```diff
@@ -1,4 +1,5 @@
 from sphinx_mdolab_theme.config import *
+import urllib.request
```
eirikurj, line 2: What is using this package?

doc/pyoptsparse.bib
```diff
@@ -0,0 +1,559 @@
+@article{Halila2020d,
```
eirikurj, line 1: Is the idea here to collect all publications that have used this particular package, or just MACH-Aero from the lab? Current the list is not exactly complete.

doc/conf.py
nwu63: Whoops I had a previous version that was using this to fetch `mdolab.bib` but eventually decided to maintain a separate file, since many references don't exist in mdolab.bib. Will remove.

doc/pyoptsparse.bib
nwu63: The idea was to collect a non-exhaustive list of papers that have used pyOptSparse. I have opted to neglect many papers from the lab since most of them use this so the list would become too long, but it was not done in a systematic fashion---I basically picked some representative MACH papers with a variety of applications (BLI, transition, strut-braced wing, etc.). Do you think a more complete list would be beneficial?

marcomangano (approved):

doc/pyoptsparse.bib
eirikurj: No, just curious. I am not sure if a complete listing here is beneficial, or feasible for that matter, as it will become outdated quickly since its currently manually updated. If we want that we could look into making a more automatically generated list by adding keywords to the original bib file. Not suggesting that for this PR though.  
If we want to highlight the breadth of works it select publication i.e., different frameworks or application, problem size etc., then we should just pick few representative papers (similar to you have). 
Also, I think we could include any Aerostructural work here as well under "MACH" or similar even though its not open-source, since those paper include some of the largest optimization that this group does with this package. Perhaps Prof. Martins will want to add specific papers here?
Its really a shame that we did not have (until recently) a separate paper what we could have cited specifically for pyOptSparse over pyOpt for older publication. It would have been easy to include a link to google scholar (or similar service) [for pyOpt](https://scholar.google.com/scholar?cites=14151378984480309081&as_sdt=2005&sciodt=0,5&hl=en) for a large amount of the publication. Maybe we should still do that and add a link to the [new one](https://scholar.google.com/scholar?cites=13781283857839029766&as_sdt=2005&sciodt=0,5&hl=en)?

nwu63: Yes exactly, and given how many papers from the lab uses pyOptSparse I'm inclined not to include every paper here. I had considered automated bib file process but the problem remains -- I don't want to just blanket include papers. See for example what OpenMDAO does on [their site](http://openmdao.org/publications/). I'll go through the list of papers here and clean up a little bit, and add some more aerostructural references. Other than that is there anything else you think I should do?

Nah that's okay I'm not really looking for a complete listing, but just some papers to highlight the wide-ranging applications.

marcomangano (approved):

doc/pyoptsparse.bib
eirikurj: I think we should probably include some of the labs earlier more cited work like, like something from Gaetan, Graeme, or Peter work, e.g., "Multipoint high-fidelity aerostructural optimization of a transport aircraft configuration", and "Aerodynamic shape optimization investigations of the common research model wing benchmark" or something similar. Maybe Prof. Martins wants to highlight specific papers? We can also just update it later.

eirikurj (approved):

doc/pyoptsparse.bib
nwu63: Neither paper mentions pyOptSparse specifically although my understanding is they used the sparse interface. If you think that's still fine to cite then I will update this PR with that and call it a day I think.

eirikurj: Yeah, they dont mention it explicitly, although I do think they probably did use it. Having said that we can just update this later.

eirikurj (approved):

marcomangano (approved):

PR #197, opened 2021-01-28
Fixed SNOPT workspace lengths

### LINKED ISSUE #196: pySNOPT is broken with MACH tools
## Description
Somehow, SNOPT doesn't work anymore when using it for MACH optimizations, despite all the analytic tests in pyOptSparse passing. This is a strange one...first reported by @friedenhe for DAfoam and reproduced by myself and @marcomangano for ADflow. Unfortunately I do not have time to work on this so any help would be appreciated.

### Steps to reproduce issue

1. Run for example `aero_opt.py` in the mdolab/MACH-Aero repo. Make sure to set the optimizer to SNOPT.

### Current behavior
The following error message is shown immediately:
```
  File "aero_opt.py", line 227, in <module>
    sol = opt(optProb, MP.sens, storeHistory=os.path.join(outputDirectory, "opt.hst"))
  File "pyoptsparse/pyoptsparse/pySNOPT/pySNOPT.py", line 427, in __call__
    self.metadata["endTime"] = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
AttributeError: 'SNOPT' object has no attribute 'metadata'
```

### Expected behavior
Optimization should complete.

### Preliminary analysis
For some reason, SNOPT seems to exit without performing _any_ analysis whatsoever. Because `masterfunc` is not called, the `metadata` field is not set and therefore the error is shown. I have verified that the SNOPT library is indeed called and `snkerc` is executed.

I believe this bug is introduced in v2.4.0, almost certainly due to the various changes to baseclasses. The most likely scenario is that we have messed up setting some SNOPT options, maybe due to the default value changes in pySNOPT.

### Possible debugging process
Check the SNOPT options in the print out file, and compare between v2.4.0 and a previous working version. That's probably a good starting point to see what options were messed up after #193. To dig deeper in to this, you can pickle all the inputs to the `snkerc` call and compare them between versions, since that's the stuff that would affect SNOPT behaviour.

Once fixed, we should also add more tests to try and capture this.

### Code versions
_List versions only if relevant_
- master branch of pyOptSparse
- SNOPT 7.7.5
- latest master branch of MACH-Aero tools

sseraj:
## Purpose
This PR closes #196 by fixing how the workspace lengths are set in pySNOPT. Prior to #193, `"Total integer workspace"` and `"Total real workspace"` were [stored in `self.options`](https://github.com/mdolab/pyoptsparse/blob/80ef5694097ff5c93026f9ccd93cc9fdee09d3b1/pyoptsparse/pySNOPT/pySNOPT.py#L417-L418). However, `_set_snopt_options` [looped over `self.set_options`](
https://github.com/mdolab/pyoptsparse/blob/80ef5694097ff5c93026f9ccd93cc9fdee09d3b1/pyoptsparse/pySNOPT/pySNOPT.py#L671), so the workspace options were not actually used.

The BaseSolver update in #193 made it so that `_set_snopt_options` [loops over `self.options`](https://github.com/mdolab/pyoptsparse/blob/master/pyoptsparse/pySNOPT/pySNOPT.py#L590) instead, so the workspace options are set. This is only problematic when the lengths are [overwritten](https://github.com/mdolab/pyoptsparse/blob/master/pyoptsparse/pySNOPT/pySNOPT.py#L358-L368) by suggested minimum values from SNOPT but the options dictionary is not updated. This explains why the analytical problems in the tests work but the more memory intensive MACH-Aero wing does not.

Adding calls to `setOption` after the lengths are overwritten would fix the bug. I instead chose to remove the initial `setOption` lines because this is closest to the behavior before #193. If we want the workspace lengths to be options that the user can actually specify, then we need to remove the [hard-coded values](https://github.com/mdolab/pyoptsparse/blob/master/pyoptsparse/pySNOPT/pySNOPT.py#L332-L335), add defaults, and rework that part of the code with `getOption` and `setOption`.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Follow the steps in #196.

## Checklist


marcomangano (approved):
Excellent detective work!!! 
This explains why I was not catching the same error on the other smaller cases I was using - my meshes were too small.

Some consideration for discussion / my understanding:
1. I see that the `Total ...` options allow to set a limit on the available workspace, but then you correctly point out that SNOPT overwrites them. Was the issue triggered here because the initial case was very big, so SNOPT automatically increases the workspace size on startup?
Or can this overwriting happen multiple times during the optimization? It just bugs me that you can potentially set an upper limit, but then the software changes it anyway. I am probably missing something.
I am not sure if we want to make these options available in a latest release, @nwu63 might have a better picture about if and when it is useful to change the default values.

2. I am also curious to understand why the error came up when setting up `self.metadata`, did `pySNOPT` fail to initialize and we missed it, only to have a crash when accessing the object later in the code?


About the tests, I think this is an additional reason why we want to [have bigger regression tests](https://github.com/mdolab/uCRM/issues/4) working on MACH/MACH-aero in the near future

eirikurj:
This patch seems to resolve the issue only if the user does not set the values in the options dict.
If the user specifies any of the three it seems still broken.
The code does allow the user to specify the options but they are essentially overwritten internally. If we dont want the user to adjust these then we should not allow them to be set.
However, if we want to be able to specify them, then as @sseraj mentioned we need to check if the user did specify them using get/set option calls and if not, we could try to estimate the sizes as done the code and then update the options.  
In general if the sizes are also updated internally we need to update/set the option also since otherwise the `print` file will not display the correct values.

marcomangano:
> In general if the sizes are also updated internally we need to update/set the option also since otherwise the `print` file will not display the correct values.

I was about to push for just removing them from the dictionary, but this made me change my mind.
I guess going ahead with the refactoring is the best option.

Thanks @eirikurj for this careful check!

marcomangano (changes requested):
See @eirikurj comment above. We might want to refactor the code to correctly handle these internally updated options

sseraj:
Thanks for the comments. I will go ahead with the refactor.

sseraj:
> 1. I see that the `Total ...` options allow to set a limit on the available workspace, but then you correctly point out that SNOPT overwrites them. Was the issue triggered here because the initial case was very big, so SNOPT automatically increases the workspace size on startup?
>    Or can this overwriting happen multiple times during the optimization? It just bugs me that you can potentially set an upper limit, but then the software changes it anyway. I am probably missing something.

SNOPT itself is not changing the limits. This is a feature in pySNOPT to essentially automate the memory allocation. This only happens once before the optimization starts.

> This explains why I was not catching the same error on the other smaller cases I was using - my meshes were too small.

From my limited understanding, it is not related to the size of the mesh but to the number and "density" of the DVs. For example, the MACH-Aero wing without shape DVs does not trigger this error.

> 2. I am also curious to understand why the error came up when setting up `self.metadata`, did `pySNOPT` fail to initialize and we missed it, only to have a crash when accessing the object later in the code?

@nwu63 points this out in #196. SNOPT exits without performing any analysis, so `self.metadata` is never defined. If you comment out the lines in Python that raise the error, you will see that SNOPT exits with `82: not enough character storage`.

nwu63:
> SNOPT itself is not changing the limits. This is a feature in pySNOPT to essentially automate the memory allocation. This only happens once before the optimization starts.

We may want to add this to the documentation, at least explaining how these are handled and whether the user can change these values.

> From my limited understanding, it is not related to the size of the mesh but to the number and "density" of the DVs. For example, the MACH-Aero wing without shape DVs does not trigger this error.

Correct, memory usage in SNOPT is dependent on many things such as the size of the optimization problem, density of the Hessian, etc. I'm curious if with sufficient large `N` in the large_sparse test we could catch this.

 
> @nwu63 points this out in #196. SNOPT exits without performing any analysis, so `self.metadata` is never defined. If you comment out the lines in Python that raise the error, you will see that SNOPT exits with `82: not enough character storage`.

Yeah this shows some bad coding on my part. Obviously `metadata` should be initialized as an empty dictionary or something in the `init`, so that the correct SNOPT error code gets propagated to the user. We should definitely fix this.

marcomangano:
> SNOPT itself is not changing the limits. This is a feature in pySNOPT to essentially automate the memory allocation. This only happens once before the optimization starts.

This makes sense, thx. +1 for adding to the documentation.

> From my limited understanding, it is not related to the size of the mesh but to the number and "density" of the DVs. For example, the MACH-Aero wing without shape DVs does not trigger this error.

Also this sounds correct, I tripped in my mind but I intended a small case from an opt, and not an analysis point of view. Probably worth trying to catch this with a test.

> @nwu63 points this out in #196. SNOPT exits without performing any analysis, so `self.metadata` is never defined. If you comment out the lines in Python that raise the error, you will see that SNOPT exits with `82: not enough character storage`.

Good check, this was what I was referring too, did not realize that SNOPT "gracefully" exits from this memory allocation issue so you don't really have an early crash. I think we can fix the `metadata` issue as Neil suggests.

marcomangano (approved):
I tested the changes locally and they work!
We might want to add that the workspace size has to be >=500 in the docs, although SNOPT exit message states it explicitly. Up to you!

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -330,12 +333,17 @@ def __call__(
...
-
-            self.setOption("Total integer workspace", leniw)
-            self.setOption("Total real workspace", lenrw)
+            lencw = self.getOption("Total character workspace")
+            leniw = self.getOption("Total integer workspace")
+            lenrw = self.getOption("Total real workspace")
+
+            # Set defaults
+            if leniw is None:
+                leniw = 500 + 100 * (ncon + nvar)
```
eirikurj, line 342: Can we define this magic number 500 anywhere, or better, connect it with and replace it with `lencw` in case that gets updated from the interface. Same below.

eirikurj (changes requested):

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -54,6 +54,9 @@ def __init__(self, raiseError=True, options={}):
             "Start": [str, ["Cold", "Warm"]],  # used in call to snkerc, the option "Cold Start" etc are NOT allowed
             "Derivative level": [int, 3],
             "Proximal iterations limit": [int, 10000],  # very large # to solve proximal point problem to optimality
+            "Total character workspace": [int, 500],  # lencw: 500
+            "Total integer workspace": [int, None],  # leniw: 500 + 100 * (m+n)
+            "Total real workspace": [int, None],  # lenrw: 500 + 200 * (m+n)
```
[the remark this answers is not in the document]
marcomangano: Agree on the comment change, I would say that the explanation might be too much for our docs. Maybe we can point to a section in the SNOPT manual?

pyoptsparse/pySNOPT/pySNOPT.py
sseraj: 500 is the minimum workspace length. SNOPT will exit with `81` if any of the lengths are lower than that. I guess I could define a variable for that? It's not really connected to `lencw`.

eirikurj: Right, exactly its the minimum for all the work arrays. My comment was more to make it more explicit that this number is the minimum in the SNOPT code itself, i.e. remind the reader of the python code why its 500, not just a random number. I suggest we at least add a comment on what this number is. Alternatively, we can also define a variable `minWorkArraySize = 500` or something.  
And you are right its not connected and they dont have to be connected. `lencw` is the only variable out of the three in the options with a default value set. I imagined that in case you change this number its likely that you need to modify the others too, so by making this connected you would essentially only need to change one instead of all three. 
However, its probably best *not* to make them connected as it better reflects the source code, and if you are going to change these numbers yourself anyways you probably have a good reason to do so and will likely need to modify all three.

sseraj:
After a discussion with @nwu63 and @marcomangano, we decided to:
- skip overwriting the lengths if the user specifies any of the lengths as an option
- add a test based on a dense version of the large sparse problem

marcomangano (approved):
Looks good to me!
The logic to trigger the workspace size update makes sense and is consistent with what we discussed. We trigger the update only if no options are prescribed by the user.

Nice touch with the test you added! That now should cover all the "branches", right? I mean 1) workspace size is set by default and it is enough for the problem (current) 2) the initial size guess is not enough for the dense problem, so SNOPT updates it automatically, re-initializes the opt and goes on with the optimization, and 3) User set values, not updated, that will fail with 80-84 in this case.

eirikurj (approved):

sseraj:
> Nice touch with the test you added! That now should cover all the "branches", right? I mean 1) workspace size is set by default and it is enough for the problem (current) 2) the initial size guess is not enough for the dense problem, so SNOPT updates it automatically, re-initializes the opt and goes on with the optimization, and 3) User set values, not updated, that will fail with 80-84 in this case.

Yes, that's right. The third branch also tests the `self.metadata` change by requiring SNOPT to exit properly.

test/test_snopt_workspace.py
```diff
@@ -0,0 +1,122 @@
...
+        test_name = "SNOPT_workspace_user"
+        optOptions = {
+            "Print file": "{}.out".format(test_name),
+            "Summary file": "{}_summary.out".format(test_name),
+            "Total character workspace": 500,
+        }
+        sol = self.optimize("snopt", optOptions=optOptions)
+
+        # Check that the workspace is too small without overwriting the lengths
+        self.assertEqual(sol.optInform["value"], 84)
```
nwu63, line 118: IIRC 84 = "not enough real storage", why is that the exit code when you've set total character to 500?

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -328,14 +331,29 @@ def __call__(
...
-            lenrw = 500 + 200 * (ncon + nvar)
+            lencw = self.getOption("Total character workspace")
+            leniw = self.getOption("Total integer workspace")
+            lenrw = self.getOption("Total real workspace")
 
-            self.setOption("Total integer workspace", leniw)
-            self.setOption("Total real workspace", lenrw)
+            # Set a flag to avoid overwriting user-specified lengths
+            if lencw is None and leniw is None and lenrw is None:
+                defaultLengths = True
```
nwu63, line 342: I would prefer a more descriptive variable name such as `checkLengths`, since it's possible for some of these to use default values while others are set by the user.

pyoptsparse/pySLSQP/pySLSQP.py
```diff
@@ -257,7 +257,7 @@ def slgrad(m, me, la, n, f, g, df, dg, x):
             self.optProb.comm.bcast(-1, root=0)
 
             # Store Results
-            inform = np.asscalar(mode)
+            inform = mode.item()
```
nwu63, line 260: What was the issue here that required the change? Did this mean all the informs were broken previously?

test/test_snopt_workspace.py
```diff
@@ -0,0 +1,122 @@
+"""
+This example is taken from the OpenOpt Examples website.
```
nwu63, line 2: Is this the same as the large_sparse but with dense Jacobians? Can you generalize that file instead of making a new file with (mostly) duplicate code for the setup portion?

nwu63 (changes requested):

test/test_snopt_workspace.py
sseraj: Setting `Total character workspace` skips the length check for all lengths. The real workspace ends up being too small. Is this the desired behavior? We could do the checks individually.

pyoptsparse/pySLSQP/pySLSQP.py
sseraj: Nothing was broken, but the tests were printing deprecation warnings.

test/test_snopt_workspace.py
sseraj: Yes, I'll do this.

test/test_snopt_workspace.py
nwu63: Yeah I think the desired behaviour is to override the automatic process individually, since the three options are completely separate.

pyoptsparse/pySLSQP/pySLSQP.py
nwu63: I see this is the `asscalar` deprecation warning.

nwu63 (approved):

test/test_large_snopt.py
```diff
@@ -0,0 +1,140 @@
...
+        }
+        self.optimize(tol=1e-5, optOptions=optOptions)
+
+    def test_default(self):
+        test_name = "SNOPT_workspace_default"
+        optOptions = {
+            "Print file": "{}.out".format(test_name),
+            "Summary file": "{}_summary.out".format(test_name),
+        }
+        sol = self.optimize(sparse=False, optOptions=optOptions)
```
sseraj, line 119: If you add `tol` here you can remove the asserts.

sseraj (commented):
Nice refactoring

PR #198, opened 2021-01-29
Changing the SLSQP output file id

eirikurj:
## Purpose
Closes https://github.com/mdolab/MACH-Aero/issues/31. 
File id 6 is often associated with `stdout` and as such was redirecting the adflow output to this file.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Run the tutorial as described in https://github.com/mdolab/MACH-Aero/issues/31.

## Checklist

### DIFF
--- a/pyoptsparse/pySLSQP/pySLSQP.py
+++ b/pyoptsparse/pySLSQP/pySLSQP.py
@@ -44,7 +44,7 @@ class SLSQP(Optimizer):
             "ACC": [float, 1e-6],  # Convergence Accurancy
             "MAXIT": [int, 500],  # Maximum Iterations
             "IPRINT": [int, 1],  # Output Level (<0 - None, 0 - Screen, 1 - File)
-            "IOUT": [int, 6],  # Output Unit Number
+            "IOUT": [int, 60],  # Output Unit Number
             "IFILE": [str, "SLSQP.out"],  # Output File Name
         }
         self.informs = {


marcomangano (approved):
Nice, finally! Can't test it rn but as long as you can get ADflow iters on the terminal I am happy with this.

nwu63 (approved):
Can confirm this worked as expected with MACH-Aero. Is the unit number 6 something that was hard-coded in ADflow? If it's general for `stdout` why did it not redirect Python outputs?

eirikurj:
No not hardcoded. In Fortran assigning 6 to stdout is not a standard but more like a convention (and a very widely used one). This can be changed in code or on a compiler level. Here * directs to 6 (the default for gfortran), i.e. when adflow calls `write(*,*)` it is the same as calling `write(6,*)`. Python follows the C convention and uses 1.

nwu63:
> No not hardcoded. In Fortran assigning 6 to stdout is not a standard but more like a convention (and a very widely used one). This can be changed in code or on a compiler level. Here * directs to 6 (the default for gfortran), i.e. when adflow calls `write(*,*)` it is the same as calling `write(6,*)`. Python follows the C convention and uses 1.

I see, so Fortran happens to use 6 and there was a clash between SLSQP and ADflow? I guess for SNOPT we use different unit numbers for print and summary files so I guess that's why it's not an issue there. THanks.

PR #205, opened 2021-02-23
Removed extra f2py wrapper files

nwu63:
## Purpose
As far as I can tell, these files are auto-generated by `f2py` during compilation, and were committed into the repo by accident. This PR removes these unnecessary files.

I have also removed the dash test in python since it was poorly designed, and also wasn't really testing anything anyways.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
All existing tests should pass.

## Checklist


marcomangano (approved):

nwu63:
The dash test was poorly designed -- the subprocess doesn't get killed cleanly so repeated execution will cause subsequent tests to fail. I've removed the test for the time being.

sseraj:
Where did the issue with the dash test show up? All the tests seem to have passed before the dash test was removed.

nwu63:
> Where did the issue with the dash test show up? All the tests seem to have passed before the dash test was removed.

It's fine on Azure, but try it locally -- the server doesn't get killed properly. So if you run `testflo` again the dash test will fail the second time.

sseraj:
> It's fine on Azure, but try it locally -- the server doesn't get killed properly. So if you run `testflo` again the dash test will fail the second time.

It's actually working fine for me. Maybe @marcomangano can check as well.

marcomangano:
> It's actually working fine for me. Maybe @marcomangano can check as well.

I think this is exactly the kind of problem I mentioned before in our chat, the second time I run the tests `test_optview_dash` fails with `AssertionError: TimeoutExpired not raised` and testflo stalls after all the other tests are run.

sseraj (approved):
Interesting. It's probably best to remove the dash test then.

marcomangano (approved):

PR #212, opened 2021-03-11
Set the gradient function "sens" at initialization of Optimization class

gjkennedy:
## Purpose

This pull request adds an optional sens argument set on initialization of the Optimization problem class. It makes sense that you should be able to set the gradient callback function at the same place you set the objective/constraint function. This makes it easier to pass around an Optimization problem class where you don't also have to keep track of the gradient function as well.

The optional argument is overridden by any value passed through the normal optimization initialization. 

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

## Checklist


nwu63 (approved):
This makes sense. I've just changed one test to use this API so this is at least being tested somewhere.

marcomangano (approved):

PR #214, opened 2021-03-13
Extract and store SNOPT version in history file

nwu63:
## Purpose
I added another key to `metadata` called `optVersion`, currently defaulting to `None`. I then wrapped the `snTitle` subroutine to extract the version information from SNOPT, and set it inside Python. This allows the history file to store the SNOPT version information which is useful for reproducibility of results.

Note that docs build will probably fail. I will update the way the options tables are generated and fix this.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I added a new test.

## Checklist

### DIFF
--- a/doc/api/history.rst
+++ b/doc/api/history.rst
@@ -14,7 +14,8 @@ In this case, the history file would have the following layout::
     │   ├── endTime
     │   ├── optTime
     │   ├── optimizer
-    │   └── version
+    │   ├── version
+    │   └── optVersion
     ├── optProb
     ├── varInfo
     │   └── xvars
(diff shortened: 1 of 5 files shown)


nwu63:
The fix for docs is a bit hacky. In the long term I would like to stop Sphinx from importing the whole module, and move to the `@staticmethod` approach used in the rest of the MACH tools. I will create an issue.

sseraj (approved):
I checked locally that this works with SNOPT 7.7.5

marcomangano (approved):
This is a nice addition! However, I agree we want to figure out a better way to handle options (and testing)

PR #219, opened 2021-03-20
Fix several issues with hotstart

### LINKED ISSUE #208: Wrong handling of hot-start optimization if a non-existing history file is provided
## Description
If a wrong `hst` file is provided as input for an hot start, pyOptSparse will fail because it will try to initialize the hot start anyway - see the error message below.

### Steps to reproduce issue
Try to run a hot-start optimization using a mis-spelled or non-existing `.hst` file

### Current behavior
The following error is printed out:
```
+------------------------------------------------------------------------------+
| pyOptSparse Warning: Hot start file does not exist. Performing a regular     |
| start                                                                        |
+------------------------------------------------------------------------------+
  File "aero_DVtest.py", line 267, in <module>
    sol = opt(optProb, MP.sens, storeHistory=histFile, hotStart=args.hotstart, timeLimit=args.timelimit)
  File "/.../miniconda3/envs/adflow-test/lib/python3.7/site-packages/pyoptsparse/pySNOPT/pySNOPT.py", line 436, in __call__
    self._setHistory(storeHistory, hotStart)
  File "/.../miniconda3/envs/adflow-test/lib/python3.7/site-packages/pyoptsparse/pyOpt_optimizer.py", line 174, in _setHistory
    val = self.hotStart.read(key)
AttributeError: 'NoneType' object has no attribute 'read'
```

### Expected behavior
As per printed message, the OptSparse should just roll back to a cold start and run a regular optimization.
NOTE: this expected behaviour should be included into the documentation!

### Code versions
- pyOptSparse 2.4.1

### LINKED ISSUE #209: Improve hotstart feature
# Description
- [ ] Document how it works
- [ ] Have it automatically set the DV to the start of the hotstart file

nwu63:
## Purpose
This PR addresses several issues with hotstart
- closes #208
- closes #209

In this process, we've had to make the following changes:
- move `setHistory` to outside of the comm split for all optimizers. This allows us to set the initial DVs into the optProb before it's parsed to generate the numpy arrays for each optimizer
- wrap `setHistory` with MPI to ensure only a single proc is reading/writing to the history file
- updated `getValues` and the History class in general to allow accessing iterations which are not analysis. For some reason, certain optimizers (IPOPT in this case) will have the initial evaluation be a gradient one, I have no idea why. Since we only want the initial DVs this doesn't matter, but the code had to be updated to handle this.

Remaining TODOs:
- [ ] add documentation on how hotstart works (I will work with @marcomangano to address this)

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
We added new tests starting from perturbed DVs, and ensuring that the hotstart completed successfully without us having to manually set the DVs.

## Checklist

### DIFF
--- a/doc/advancedFeatures.rst
+++ b/doc/advancedFeatures.rst
@@ -28,11 +28,10 @@ For this process to work, the following must be true:
    If the optimizer performs a random gradient check (e.g. SNOPT), it's best to disable these just in case.
 -  Optimizer settings that affect the path of the optimization must remain the same.
    It is perfectly fine to change for example settings related to print outs, but not those affecting the line search.
--  The hot start file is itself cold started.
-   It is not possible to hot start from another hot start.
 
 To use the hot start feature, simply call the optimizer with the option ``hotStart = <hot start file>``.
 See the API documentation for each optimizer for more information.
+Because the hot start process will store all the previous "restarted" iterations in the new history file, it's possible to restart as many times as you like, each time using the previous history file.
 
 
 
--- a/doc/advancedFeatures.rst
+++ b/doc/advancedFeatures.rst
@@ -10,7 +10,7 @@ Advanced Features
 .. ----------------------------
 
 Hot start
---------
+---------
 Hot start refers to the way optimizations are initialized.
 Suppose you run an optimization, and it was accidentally terminated prematurely, for example due to an iteration limit which was too low.
 If you restarted the optimization from the last iteration, you will likely incur a performance penalty, and end up taking far more steps than it would've taken if the original optimization was allowed to progress.
@@ -24,9 +24,12 @@ If at any point the requested design variables diverge from the history, then we
 For this process to work, the following must be true:
 
 -  The optimizer is deterministic.
-   Note that a stochastic optimizer is still deterministic even if it uses random numbers, as long as the initial seed is fixed.
+   Note that a stochastic optimizer can still be made deterministic even if it uses random numbers, as long as the random seed is fixed.
+   If the optimizer performs a random gradient check (e.g. SNOPT), it's best to disable these just in case.
 -  Optimizer settings that affect the path of the optimization must remain the same.
    It is perfectly fine to change for example settings related to print outs, but not those affecting the line search.
+-  The hot start file is itself cold started.
+   It is not possible to hot start from another hot start.
 
 To use the hot start feature, simply call the optimizer with the option ``hotStart = <hot start file>``.
 See the API documentation for each optimizer for more information.
--- a/doc/advancedFeatures.rst
+++ b/doc/advancedFeatures.rst
@@ -12,14 +12,14 @@ Advanced Features
 Hot start
 ---------
 Hot start refers to the way optimizations are initialized.
-Such start-up procedure is named in contrast to the regular optimization intialization, or "cold start", in which case pyOptSparse simply initializes an optimization job from scratch, using an arbitrary, user-defined set of initial design variables.
-There are several situations in which such cold optimization starts can be avoided by leveraging on the information from a previous optimization, with the aim to reduce the overall computational time.
+Such start-up procedure is named in contrast to the regular optimization initialization, or "cold start", in which case pyOptSparse simply initializes an optimization job from scratch, using initial design variables set by the user.
+There are several situations in which such cold starts can be avoided by leveraging on the information from a previous optimization, with the aim to reduce the overall computational time.
 Suppose you run an optimization that was accidentally terminated prematurely by, for example, an excessively-low iteration limit.
 If you restarted the optimization using the DVs from the last iteration (but losing all the accumulated history), you will start from a better initial point but will likely incur in a performance penalty.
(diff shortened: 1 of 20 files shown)


marcomangano:
Interestingly enough, `test_hs015.py:TestHS15.test_conmin`, `test_hs015.py:TestHS15.test_slsqp` fail by a relatively small discrepancy, `Max relative difference: 0.00691044` and `2.11013994e-07` respectively. I guess this depends on the new optimization path.. is there a limit in iterations we are hitting? or an artificial local minima in `hs015`?

nwu63:
We did the following to fix the tests:
1. start the hotstart optimization with different DVs at a point which is close to the original starting point, which will help certain optimizers (SLSQP and CONMIN) converge to the final optimum. Practically speaking, the tolerances on these optimizers are a bit too tight
2. loosen the PSQP tolerance slightly
3. removed SLSQP from the list of optimizers which guarantee that `last` will contain funcs, since that does not appear to be the case.

I will start working on the docs now

test/test_hs015.py
```diff
@@ -176,7 +176,8 @@ def check_hist_file(self, optimizer, tol):
         val = hist.getValues()
 
         # this check is only used for optimizers that guarantee '0' and 'last' contain funcs
-        if optimizer in ["SNOPT", "SLSQP", "PSQP"]:
+        if optimizer in ["SNOPT", "PSQP"]:
+            print(hist.read("last"))
```
marcomangano, line 180: Debugging printout?

nwu63: Oh yes good catch, will fix

marcomangano: So, we just skip SLSQP check there? What's wrong with the restart?

nwu63: My understanding is that SLSQP does not guarantee that the last call counter contains `funcs`, so I have taken it out of the list of tests here.

sseraj:
How did the tests pass before "minor update" and then fail only on two images after an rst file was changed? Are the tests non-deterministic?

nwu63:
I have absolutely no idea.. somehow `mpi4py` was not available on the image? It's really strange.

That being said, we should probably explicitly test the case without `mpi4py` too.

doc/advancedFeatures.rst
```diff
@@ -0,0 +1,43 @@
...
+If at any point the requested design variables diverge from the history, then we will actually perform a function evaluation, and all subsequent points will be newly evaluated.
+For this process to work, the following must be true:
+
+-  The optimizer is deterministic.
+   Note that a stochastic optimizer can still be made deterministic even if it uses random numbers, as long as the random seed is fixed.
+   If the optimizer performs a random gradient check (e.g. SNOPT), it's best to disable these just in case.
+-  Optimizer settings that affect the path of the optimization must remain the same.
+   It is perfectly fine to change for example settings related to print outs, but not those affecting the line search.
+-  The hot start file is itself cold started.
+   It is not possible to hot start from another hot start.
```
marcomangano, line 32: Can you expand on this a bit? From this it sounds like that you can hot-start an optimization only once, but I am pretty sure this is not the case, right?

doc/advancedFeatures.rst
```diff
@@ -0,0 +1,43 @@
...
+
+.. Parallel Execution
+.. ------------------
+
+.. Storing Optimization History
+.. ----------------------------
+
+Hot start
+---------
+Hot start refers to the way optimizations are initialized.
```
marcomangano, line 14: I can add a one-liner about the difference with "cold" starts?

doc/advancedFeatures.rst
```diff
@@ -0,0 +1,43 @@
...
+Suppose you run an optimization, and it was accidentally terminated prematurely, for example due to an iteration limit which was too low.
+If you restarted the optimization from the last iteration, you will likely incur a performance penalty, and end up taking far more steps than it would've taken if the original optimization was allowed to progress.
```
marcomangano, line 16: Worth clarifying that this happens bc you restart "from scratch" with just a different set of initial DVs?
E.g. for some GB opts the first iterations are expensive bc you need to rebuild the Hessian? Also links with the following sentence

doc/advancedFeatures.rst
nwu63: Oh shoot my mistake, you are right I think I just confused myself there. Because the history file contains the full history including the previous history file, it's fine to hotstart multiple times. I will fix this.

doc/advancedFeatures.rst
nwu63: Yep good idea.

doc/advancedFeatures.rst
nwu63: Yeah that would work well. Would you mind doing some small edits?

marcomangano: Sure, I will edit this by tomorrow the latest, thanks for having done most of the job already!

marcomangano (approved):

sseraj (approved):

PR #225, opened 2021-04-02
Made stored function values real

### LINKED ISSUE #76: Make stored funcs values real instead of complex

sseraj:
## Purpose
When `objFun` is defined as a complex function to work with the complex-step derivative, there are two separate cases where complex values leak into the stored function values:

1) In the solution object. This was previously addressed in #102 but only for printing. I reverted this commit and changed `processObjtoVec` and `processContoVec` so that the stored value is real instead.

2) In the history file. I fixed this by casting `funcs` to real before they are cached.

Closes #76

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimizer.py
+++ b/pyoptsparse/pyOpt_optimizer.py
@@ -358,7 +358,12 @@ class Optimizer(BaseSolver):
 
                 self.userObjTime += time.time() - timeA
                 self.userObjCalls += 1
-                # User values stored is immediately
+
+                # Discard zero imaginary components in funcs
+                for key, val in funcs.items():
+                    funcs[key] = np.real(val)
+
+                # Store user values
                 self.cache["funcs"] = copy.deepcopy(funcs)
 
                 # Process constraints/objectives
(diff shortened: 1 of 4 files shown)


pyoptsparse/pyOpt_optimizer.py
```diff
@@ -358,7 +358,12 @@ def _masterFunc2(self, x, evaluate, writeHist=True):
 
                 self.userObjTime += time.time() - timeA
                 self.userObjCalls += 1
-                # User values stored is immediately
+
+                # Discard zero imaginary components in funcs
+                for key, val in funcs.items():
```
nwu63, line 363: Can you explain why this is needed in addition to the casting above (e.g. `con.value = np.real(copy.copy(c))`)

sseraj: The changes in the `Optimization` class only affect the values stored in `optProb.objectives` and `optProb.constraints` . Here `funcs` is coming straight from `objFun` (line 344), so it could be complex.

nwu63: Right that makes sense.

nwu63 (approved):
I think this looks fine as is, though the more I look at the code overall the more I think it's messy (for example I would have to really think about how many times `objfunc` is wrapped).

marcomangano (approved):
Good stuff, thanks for addressing this!

PR #226, opened 2021-04-03
Formatting updates + added isort

nwu63:
## Purpose
This is another formatting update, with the following changes:
- added `isort` to sort imports in a deterministic fashion (run `isort -c .` to check, and `isort .` to sort)
- removed shebang at the beginning of various files
- removed unused `if __name__ == "__main__":` blocks
- moved some shared constants to the `utils` file
	- INIFINITY (1e20)
	- EPS (renamed from eps, fetched from numpy)

I've left this as a draft PR because I'm not entirely sure we want to use `isort`. I really like the concept but I'm hesitant to enforce it. Thoughts? In any case, if we don't want some stuff in this PR I can just open a new one and cherry-pick the commits we want.

## Type of change
What types of change is it?

- [x] Code style update (formatting, renaming)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
All existing tests pass

## Checklist


pyoptsparse/pyNLPQLP/pyNLPQLP.py
```diff
@@ -14,9 +14,6 @@
 from ..pyOpt_optimizer import Optimizer
 from ..pyOpt_error import Error
 
-# constants
-eps = np.finfo(np.float64).eps
```
marcomangano, line 30: what is this?

marcomangano (commented):
Lots of interesting things here, but maybe we want to have an offline conversation about what is worth and what is not - e.g. I like `iSort` pretty much, but do we want to put the effort to test/maintain this formatting?

pyoptsparse/pyNLPQLP/pyNLPQLP.py
nwu63: This is [machine epsilon](https://en.wikipedia.org/wiki/Machine_epsilon). We use it to compare design variables during hot start, for example.

marcomangano (approved):

sseraj (commented):
Does this fix the OptView linting issues?

nwu63:
> Does this fix the OptView linting issues?

Nope, there are a bunch more and I just couldn't be bothered to fix them. I guess I could though if you really wanted.

sseraj:
No, that's alright.

pyoptsparse/pyIPOPT/pyIPOPT.py
```diff
@@ -1,38 +1,26 @@
...
 pyIPOPT - A python wrapper to the core IPOPT compiled module.
 """
-# =============================================================================
-# IPOPT Library
-# =============================================================================
-
+# Compiled module
 try:
-    from . import pyipoptcore
+    from . import pyipoptcore  # isort: skip
```
sseraj, line 6: Can you explain why these have to be skipped?

nwu63: So there are two scenarios here.
1. For what you showed here, these are not necessary. If I don't skip them, then `isort` will just add a header here:
   ```
   -    from . import snopt  # isort: skip
   +    # Local modules
   +    from . import snopt
   ```
   This is a stylistic thing, and in the future I would like to move these imports inside the `__init__` which will eliminate this issue. However, if you think these are unnecessary we can definitely remove them.
2. There are other imports which are buried deep in the file. For example [this line](https://github.com/mdolab/pyoptsparse/blob/update-format/pyoptsparse/pyOpt_history.py#L220) imports the pyOptSparse version, and adding the skip prevents `isort` from adding an import header there which would be unnecessary.

sseraj (approved):

PR #230, opened 2021-04-08
Added type hints and some minor refactoring

nwu63:
## Purpose
This is going to be a long-lived PR tracking efforts to add type hints to pyOptSparse. Nominally `mypy` has yet another configuration file, so I'm not sure if we want to include that in the repo, or distribute it separately (say via the `.github` repo as we currently do with `.flake8`).

I have also made some minor code-related changes, mainly to make mypy and flake8 happy
- Refactored the initialization of the Solution object. Timing/call counter info is now passed in via a dictionary, and we no longer manually inject class attributes arbitrarily.
- The parameter `type=` for `addVar()` and `addVarGroup()` have changed to `varType`, in order to avoid a clash with the builtin Python function `type()`. Since this is a backwards-incompatible change, we have for now allowed for both, with the value from `type=` if present overwriting the value from `varType=`. A DeprecationWarning is thrown if `type=` is used, and this will be removed in a future release.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
N/A, but `mypy` can be used to check for errors in type hinting.

## Checklist


pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -644,33 +646,6 @@ def _set_snopt_options(self, iPrint, iSumm, cw, iw, rw):
             elif isinstance(value, type(None)):
                 snopt.snset(name, iPrint, iSumm, inform, cw, iw, rw)
 
-        return
-
-    def _on_getInform(self, infocode):
```
nwu63, line 649: This code was never used, and I figured we can probably just use the super() implementation.

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -481,12 +484,11 @@ def __call__(
 
         else:  # We are not on the root process so go into waiting loop:
             self._waitLoop()
-            sol = None
 
         # Communication solution and return
-        sol = self._communicateSolution(sol)
+        commSol = self._communicateSolution(sol)
```
nwu63, line 489: I renamed these because `mypy` was getting confused and thinking `sol` could be `None`, so the answer was just to reassign this a new variable name.

pyoptsparse/pyOpt_solution.py
```diff
@@ -49,16 +66,10 @@ def __str__(self):
         text1 += "    Total Time: %25.4f\n" % self.optTime
         text1 += "       User Objective Time :   %10.4f\n" % self.userObjTime
         text1 += "       User Sensitivity Time : %10.4f\n" % self.userSensTime
-        if hasattr(self, "userJProdTime"):
```
nwu63, line 52: These were never used, probably left over from the matrix-free optimizer.

pyoptsparse/pyOpt_variable.py
```diff
@@ -26,7 +26,7 @@ def __init__(
         self.name = name
         self.type = type
         self.scalar = scalar
-        self.choices = None
+        self.choices = choices
```
marcomangano, line 29: This line is repeated on line 53, I guess we can just keep one?

marcomangano (commented):
Looking good! I am keeping an eye on this so I don't have to review it from scratch when the time comes

pyoptsparse/pyOpt_variable.py
nwu63: Oh yeah L53 should be removed.

nwu63:
I'm opening up this PR since I don't think I will be adding more into it, and it's unlike that someone else would add to this so we should probably not let it get too stale.

marcomangano (approved):
As you say, this is good to go for now. Thanks for setting this up!

pyoptsparse/pyOpt_optimization.py
```diff
@@ -137,7 +142,17 @@ def checkConName(self, conName):
             return validName
 
     def addVarGroup(
-        self, name, nVars, type="c", value=0.0, lower=None, upper=None, scale=1.0, offset=0.0, choices=None, **kwargs
+        self,
+        name: str,
+        nVars: int,
+        varType: str = "c",
```
nwu63, line 148: We changed the name of this parameter here since `type` is a built-in function. This is a backwards-incompatible change, so what we should do is check if `type` is passed in through `kwargs`, and raise warning + set `varType` to `type`. This can be removed in a future release.

sseraj: Are you planning on adding the deprecation warning to this PR?

marcomangano (changes requested):
@sseraj yes the warning will be added, updating my review to ensure this is not unintentionally merged

pyoptsparse/pyOpt_optimizer.py
```diff
@@ -758,28 +757,14 @@ def _assembleObjective(self):
 
         return np.real(np.squeeze(ff))
 
-    def _createSolution(self, optTime, sol_inform, obj, xopt, multipliers=None):
+    def _createSolution(self, optTime, sol_inform, obj, xopt, multipliers=None) -> Solution:
```
sseraj, line 760: It looks like there is some significant refactoring here that is not just type hints. Maybe we should update the PR title/description?

nwu63: I forgot about this.. I defined these class attributes explicitly in the Solution object, before we were just injecting these in random places which was pretty poor. I will update the descriptions.

pyoptsparse/pyOpt_constraint.py
```diff
@@ -23,10 +36,10 @@ def __init__(self, name, nCon, linear, wrt, jac, lower, upper, scale):
         self.linear = linear
         self.wrt = wrt
         self.jac = jac
-        self.partialReturnOk = None
+        self.partialReturnOk: Optional[bool] = None
         self.scale = scale
-        self.rs = None
-        self.re = None
+        self.rs: int
```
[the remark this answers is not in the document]
nwu63: No it does not, this just annotates the type, and they get set in `self.finalize()`. The alternative is to do something like `self.re: Optional[int] = None`, I can change it to that if you prefer.

pyoptsparse/pyOpt_solution.py
```diff
@@ -6,7 +6,7 @@
 
 
 class Solution(Optimization):
-    def __init__(self, optProb, optTime, optInform):
```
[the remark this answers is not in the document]
nwu63: True, but I am fairly certain that almost nobody is creating Solution objects outside of pyOptSparse.. I don't really consider `Solution()` part of the public API, similar to `Optimizer` classes and so on. Are you okay with this change?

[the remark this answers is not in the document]
nwu63: In any case I will update the docstrings.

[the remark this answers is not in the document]
sseraj: That's fair. The change is good with me.

pyoptsparse/pyOpt_optimization.py
nwu63: This has been added.

sseraj (approved):

pyoptsparse/pyOpt_constraint.py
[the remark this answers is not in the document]
sseraj: I missed this comment yesterday. What's the best practice here?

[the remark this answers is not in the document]
nwu63: I guess giving it the `None` value is best, I will do that.

pyoptsparse/pyOpt_optimizer.py
```diff
@@ -60,34 +62,33 @@ def __init__(
             caseSensitiveOptions=caseSensitiveOptions,
         )
         self.callCounter = 0
-        self.sens = None
-        self.optProb = None
-        self.version = version
+        self.sens: Union[None, Callable, Gradient] = None
+        self.optProb: Optimization
```
sseraj, line 66: I think this is the only other place where the initialization is missing.

nwu63: This is a bit tricky because the rest of the code relies on `self.optProb` being set, so `mypy` will complain that `NoneType` has no attribte `setDVs` etc. In the future, I would like to refactor the code which will solve these problems, but I think this is okay for now.

sseraj (approved):

marcomangano (approved):

PR #232, opened 2021-04-15
Fix counters bug

nwu63:
## Purpose
In #225, a bug was introduced whereby `self.nObj` was set to 0 in the init, and `finalize()` would increment this counter. However, multiple invocations of `finalize()` would not reset it to zero first, resulting in erroneous `nObj` values. This PR resets these counters so that `finalze()` is once again safe for multiple invocations. Note that this is only relevant if the user has changed `optProb` in between, since otherwise `finalize()` would be skipped altogether via the `self.finalized` flag check.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
None, since we never tested against this use case.

## Checklist

### DIFF
--- a/test/test_optProb.py
+++ b/test/test_optProb.py
@@ -241,6 +241,26 @@ class TestOptProb(unittest.TestCase):
                 scale = np.hstack(self.conScale)
             assert_allclose(val_opt, val_user * scale)
 
+    def test_finalize(self):
+        """
+        Check that multiple finalize calls don't mess up the optProb
+        """
+        self.setup_optProb(nObj=1, nDV=[4, 8], nCon=[2, 3], xScale=[1.0, 1.0], conScale=[1.0, 1.0], offset=[0, 0])
+        self.assert_optProb_size(1, 12, 5)
+        self.optProb.addObj("obj2")
+        self.assert_optProb_size(2, 12, 5)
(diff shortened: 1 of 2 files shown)


marcomangano (approved):
Good catch!

marcomangano:
Actually... Do we want to add a test on this?

nwu63:
Yeah I just added a test, this could be the start of some better unit tests for `addVarGroup` etc.

test/test_optProb.py
```diff
@@ -241,6 +241,26 @@ def processValue(key, val, output):
...
+        self.assert_optProb_size(1, 12, 5)
+        self.optProb.addObj("obj2")
+        self.assert_optProb_size(2, 12, 5)
+        self.optProb.addVar("DV2")
+        self.assert_optProb_size(2, 13, 5)
+        self.optProb.addCon("CON2")
+        self.assert_optProb_size(2, 13, 6)
+
+    def assert_optProb_size(self, nObj, nDV, nCon):
+        """Checks that nObj, nDV adn nCon are correct for self.optProb"""
```
sseraj, line 258: typo "adn"

sseraj (approved):

marcomangano (approved):

PR #233, opened 2021-04-16
Revert IPOPT print level to optimizer default

marcomangano:
## Purpose
Removing the default pyIPOPT `print_level: 0` option and revert to IPOPT default (`print_level: 5`). 
This provides a more verbose and useful output on both the terminal and the `.out` file.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
Verified locally by multiple users

## Checklist

### DIFF
--- a/doc/optimizers/IPOPT_options.yaml
+++ b/doc/optimizers/IPOPT_options.yaml
@@ -1,5 +1,3 @@
-print_level:
-  desc: Printing level
 output_file:
   desc: The name of the output file from IPOPT
 option_file_name:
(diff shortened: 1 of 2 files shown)


sseraj (approved):
Should we also remove this from IPOPT_options.yaml or is it okay to keep that

marcomangano:
> Should we also remove this from IPOPT_options.yaml or is it okay to keep that

Good point, I am removing it now.

sseraj (approved):

nwu63 (approved):

PR #234, opened 2021-04-17
docsFix

Xiaosong2105:
## Purpose
Explain the goal of this PR, if it addresses an existing issue be sure to link to it. Describe the big picture of your changes here, perhaps using a bullet list if multiple changes are done to accomplish a single goal. If it accomplishes multiple goals, it may be best to create separate PR's for each. 

## Type of change
What types of change is it?

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

## Checklist

### DIFF
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -205,12 +205,19 @@ Analytic Derivatives
 ++++++++++++++++++++
 
 pyOptSparse can automatically compute derivatives of the objective and constraint functions using finite differences or the complex-step method (See implementation details `here <https://mdolab-pyoptsparse.readthedocs-hosted.com/en/latest/api/gradient.html>`__).
-If analytic derivatives are available, users can define them within the ``sens()`` function.
+If analytic derivatives are available, users can define them within a user-defined function (let's use the function name ``sens()`` for the explanation purposes).
+This ``sens()`` function reads as inputs a dictionary containing design variables as well as another dictionary containing objective and constraint functions.
 Note that when supplying analytic derivatives, users need to make sure the keys in the ``sens()`` function are consistent with the keys of design variables (e.g. ``var_name``), constraints (e.g. ``equality_constraint``) and objectives (e.g. ``obj_name``).
-Based on these keys (or names) users can separately define the analytic derivatives using a nested dictionary ``funcsSens``.
+These names are simply examples used in previous paragraphs but they can be anything users prefer.
+Based on these keys (or names) users can separately define the analytic derivatives using a nested dictionary ``funcsSens`` (again, for the explanation convenience).
 Specifically, the first-layer keys should be associated with constraint and objective names while the second-layer keys are corresponding to design variables.
 The dictionary values are the provided analytic derivative lists whose column number equals the amount of objectives (or constraints) and row number equals the amount of design variables.
-Once this ``sens()`` function is constructed, users can link it to pyOptSparse by::
+For example, if the optimization problem has one objective (``obj``), two constraints (``con``), and three design variables (``xvars``), the ``funcsSens`` dictionary (assuming some simple explicit-value derivatives)  will look like::
+
+  funcsSens = { "obj": { "xvars": [1, 2, 3]  }, "con": { "xvars": [[4, 5, 6], [7, 8, 9]]  }  }
+
+The ``sens()`` function returns this ``funcsSens`` dictionary and a failure flag (most of the time ``False``) in the end.
+Once this ``sens()`` function is constructed, users can link it to `pyOptSparse <https://mdolab-pyoptsparse.readthedocs-hosted.com/en/latest/api/optimizer.html#pyoptsparse.pyOpt_optimizer.OPT>` by::
 
   sol = opt(optProb, sens=sens, ...)
 
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -216,11 +216,11 @@ If analytic derivatives are available, users can compute them within a user-defi
 This function accepts as inputs a dictionary containing design variable values as well as another dictionary containing objective and constraint values.
 It returns a nested dictionary containing the gradients of the objective and constraint values with respect to those design variables at the current design point.
 Specifically, the first-layer keys should be associated with objective and constraint names while the second-layer keys correspond to design variables.
-The dictionary values are the computed analytic derivatives, either in the form of lists or NumPy arrays whose column number equals the number of objectives (or constraints) and row number equals the number of design variables.
+The dictionary values are the computed analytic derivatives, either in the form of lists or NumPy arrays with the expected shape.
 Since pyOptSparse uses string indexing, users need to make sure the keys in the returned dictionary are consistent with the names of design variables, constraints and objectives which were first added to the optimization problem.
 
 .. tip::
-  #. Only the non-zero sub-blocks of the Jacobian need to be specified, and pyOptSparse will assume the rest to be zeros.
+  #. Only the non-zero sub-blocks of the Jacobian need to be defined in the dictionary, and pyOptSparse will assume the rest to be zero.
   #. Derivatives of the linear constraints do not need to be given here, since they are constant and should have already been specified via the ``jac=`` keyword argument when adding the constraint.
 
 For example, if the optimization problem has one objective ``obj``, two constraints ``con``, and three design variables ``xvars``, the returned sensitivity dictionary (with placeholder values) should have the following structure:
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -216,11 +216,11 @@ If analytic derivatives are available, users can compute them within a user-defi
 This function accepts as inputs a dictionary containing design variable values as well as another dictionary containing objective and constraint values.
 It returns a nested dictionary containing the gradients of the objective and constraint values with respect to those design variables at the current design point.
 Specifically, the first-layer keys should be associated with objective and constraint names while the second-layer keys correspond to design variables.
-The dictionary values are the computed analytic derivatives, either in the form of lists or NumPy arrays whose column number equals the amount of objectives (or constraints) and row number equals the amount of design variables.
+The dictionary values are the computed analytic derivatives, either in the form of lists or NumPy arrays whose column number equals the number of objectives (or constraints) and row number equals the number of design variables.
 Since pyOptSparse uses string indexing, users need to make sure the keys in the returned dictionary are consistent with the names of design variables, constraints and objectives which were first added to the optimization problem.
 
 .. tip::
-  #. Only the non-zero sub-blocks of the Jacobian needs to be specified, and pyOptSparse will assume the rest to be zeros.
+  #. Only the non-zero sub-blocks of the Jacobian need to be specified, and pyOptSparse will assume the rest to be zeros.
   #. Derivatives of the linear constraints do not need to be given here, since they are constant and should have already been specified via the ``jac=`` keyword argument when adding the constraint.
 
 For example, if the optimization problem has one objective ``obj``, two constraints ``con``, and three design variables ``xvars``, the returned sensitivity dictionary (with placeholder values) should have the following structure:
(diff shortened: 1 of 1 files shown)


doc/guide.rst
```diff
@@ -201,6 +201,21 @@ What this does is tell pyOptSparse that the key ``obj_name`` in the function ret
 For optimizers that can do multi-objective optimization (e.g. NSGA2), multiple objectives can be added.
 Optimizers that can only handle one objective enforce that only a single objective is added to the optimization description.
 
+Analytic Derivatives
+++++++++++++++++++++
+
+pyOptSparse can automatically compute derivatives of the objective and constraint functions using finite differences or the complex-step method (See implementation details `here <https://mdolab-pyoptsparse.readthedocs-hosted.com/en/latest/api/gradient.html>`__).
+If analytic derivatives are available, users can define them within the ``sens()`` function.
```
marcomangano, line 208: I think `sens()` can have any name, it's the `sens=` key in the `Optimizer` instance that has to be used, as you point out a few lines later. Just specify that `sens` is only for explanation purposes here.

doc/guide.rst
```diff
@@ -201,6 +201,21 @@ What this does is tell pyOptSparse that the key ``obj_name`` in the function ret
...
+Note that when supplying analytic derivatives, users need to make sure the keys in the ``sens()`` function are consistent with the keys of design variables (e.g. ``var_name``), constraints (e.g. ``equality_constraint``) and objectives (e.g. ``obj_name``).
```
marcomangano, line 209: I would specify that the example names come from the previous paragraph, I was getting confused and other people might not read all the docs every time :)

doc/guide.rst
```diff
@@ -201,6 +201,21 @@ What this does is tell pyOptSparse that the key ``obj_name`` in the function ret
...
+Based on these keys (or names) users can separately define the analytic derivatives using a nested dictionary ``funcsSens``.
```
marcomangano, line 210: We should explicitly provide an example of the `funcSens` dictionary (again, naming is flexible) as it would really facilitate the user understanding. You also did not mention the input arguments of the `sens(x, funcs)` function,  which again have a specific signature - please add it.

doc/guide.rst
```diff
@@ -201,6 +201,21 @@ What this does is tell pyOptSparse that the key ``obj_name`` in the function ret
...
+Specifically, the first-layer keys should be associated with constraint and objective names while the second-layer keys are corresponding to design variables.
+The dictionary values are the provided analytic derivative lists whose column number equals the amount of objectives (or constraints) and row number equals the amount of design variables.
+Once this ``sens()`` function is constructed, users can link it to pyOptSparse by::
+
+  sol = opt(optProb, sens=sens, ...)
+
+This ``opt`` function is an optimizer instantiation which comes in the following section.
```
marcomangano, line 217: I would add a :class: reference for Optimizer.

marcomangano (changes requested):
Thanks for working on this! I have a few recommended edits and I am sure @nwu63 will have more detailed comments

doc/guide.rst
Xiaosong2105: Good point. I will update this.

doc/guide.rst
Xiaosong2105: Sure, will do.

doc/guide.rst
Xiaosong2105: No problem. I will add it.

doc/guide.rst
Xiaosong2105: No problem.

Xiaosong2105:
> Thanks for working on this! I have a few recommended edits and I am sure @nwu63 will have more detailed comments

Thanks Marco for looking into this. I have made further changes based on your comments. Please check.

doc/guide.rst
```diff
@@ -201,6 +201,28 @@ What this does is tell pyOptSparse that the key ``obj_name`` in the function ret
 For optimizers that can do multi-objective optimization (e.g. NSGA2), multiple objectives can be added.
 Optimizers that can only handle one objective enforce that only a single objective is added to the optimization description.
 
+Analytic Derivatives
+++++++++++++++++++++
+
+pyOptSparse can automatically compute derivatives of the objective and constraint functions using finite differences or the complex-step method (See implementation details `here <https://mdolab-pyoptsparse.readthedocs-hosted.com/en/latest/api/gradient.html>`__).
```
nwu63, line 207: These can all be replaced by internal Sphinx links, e.g. :ref:`here <gradient>` (I think..)

doc/guide.rst
```diff
@@ -201,6 +201,28 @@ What this does is tell pyOptSparse that the key ``obj_name`` in the function ret
...
+If analytic derivatives are available, users can define them within a user-defined function (let's use the function name ``sens()`` for the explanation purposes).
+This ``sens()`` function reads as inputs a dictionary containing design variables as well as another dictionary containing objective and constraint functions.
+Note that when supplying analytic derivatives, users need to make sure the keys in the ``sens()`` function are consistent with the keys of design variables (e.g. ``var_name``), constraints (e.g. ``equality_constraint``) and objectives (e.g. ``obj_name``).
+These names are simply examples used in previous paragraphs but they can be anything users prefer.
+Based on these keys (or names) users can separately define the analytic derivatives using a nested dictionary ``funcsSens`` (again, for the explanation convenience).
+Specifically, the first-layer keys should be associated with constraint and objective names while the second-layer keys are corresponding to design variables.
+The dictionary values are the provided analytic derivative lists whose column number equals the amount of objectives (or constraints) and row number equals the amount of design variables.
```
nwu63, line 214: They're not lists they are usually numpy arrays.

doc/guide.rst
```diff
@@ -201,6 +201,28 @@ What this does is tell pyOptSparse that the key ``obj_name`` in the function ret
...
+If analytic derivatives are available, users can define them within a user-defined function (let's use the function name ``sens()`` for the explanation purposes).
+This ``sens()`` function reads as inputs a dictionary containing design variables as well as another dictionary containing objective and constraint functions.
+Note that when supplying analytic derivatives, users need to make sure the keys in the ``sens()`` function are consistent with the keys of design variables (e.g. ``var_name``), constraints (e.g. ``equality_constraint``) and objectives (e.g. ``obj_name``).
+These names are simply examples used in previous paragraphs but they can be anything users prefer.
+Based on these keys (or names) users can separately define the analytic derivatives using a nested dictionary ``funcsSens`` (again, for the explanation convenience).
+Specifically, the first-layer keys should be associated with constraint and objective names while the second-layer keys are corresponding to design variables.
```
nwu63, line 213: It's important to mention here that you only need to specify non-zero sub-blocks. If `con1` does not depend on `x1` then you can leave that part blank, and pyOptSparse will assume it's zero.

doc/guide.rst
Xiaosong2105: I think they can be lists or arrays. For example, in the pyoptsparse/examples directory, hs015.py has them as list while hs071.py has them as arrays. I will mention both of them then.

doc/guide.rst
Xiaosong2105: Oh, good to know. I just manually specify zero gradients. Will add this point.

doc/guide.rst
Xiaosong2105: I am not quite familiar with such commands. Plus, it might be good to be consistent with the other sections. So let's leave them for later if we finally decide to make such changes.

marcomangano (dismissed):

nwu63:
@sseraj can you take a look at this PR? I don't want to merge it since I contributed to this

doc/guide.rst
```diff
@@ -201,29 +201,40 @@ What this does is tell pyOptSparse that the key ``obj_name`` in the function ret
...
+~~~~~~~~~~~~~~~~~~~~
+If analytic derivatives are available, users can compute them within a user-defined function.
+This function accepts as inputs a dictionary containing design variable values as well as another dictionary containing objective and constraint values.
+It returns a nested dictionary containing the gradients of the objective and constraint values with respect to those design variables at the current design point.
+Specifically, the first-layer keys should be associated with objective and constraint names while the second-layer keys correspond to design variables.
+The dictionary values are the computed analytic derivatives, either in the form of lists or NumPy arrays whose column number equals the amount of objectives (or constraints) and row number equals the amount of design variables.
+Since pyOptSparse uses string indexing, users need to make sure the keys in the returned dictionary are consistent with the names of design variables, constraints and objectives which were first added to the optimization problem.
+
+.. tip::
+  #. Only the non-zero sub-blocks of the Jacobian needs to be specified, and pyOptSparse will assume the rest to be zeros.
```
sseraj, line 223: Should be 'need' not 'needs'

doc/guide.rst
```diff
@@ -201,29 +201,40 @@ What this does is tell pyOptSparse that the key ``obj_name`` in the function ret
...
+In the simplest case, using ``sens="FD"`` will be enough to run an optimization using forward differences with a default step size.
+
 Analytic Derivatives
-++++++++++++++++++++
+~~~~~~~~~~~~~~~~~~~~
+If analytic derivatives are available, users can compute them within a user-defined function.
+This function accepts as inputs a dictionary containing design variable values as well as another dictionary containing objective and constraint values.
+It returns a nested dictionary containing the gradients of the objective and constraint values with respect to those design variables at the current design point.
+Specifically, the first-layer keys should be associated with objective and constraint names while the second-layer keys correspond to design variables.
+The dictionary values are the computed analytic derivatives, either in the form of lists or NumPy arrays whose column number equals the amount of objectives (or constraints) and row number equals the amount of design variables.
```
sseraj, line 219: I don't think 'amount' is the right word. It should just be 'number'. 
Also, how do you specify just the non-zero terms if the matrix has to be the full size?

sseraj (commented):
We can also add `sens` through the `Optimization` object now. Is that worth mentioning?

doc/guide.rst
nwu63: It's a bit hard to explain, but it's similar to the discussion above on constraints -- you can skip sub-blocks which are nonzero altogether, but within each sub-block (i.e. each `funcsSens[output_key][input_key]`) the array shapes must match. Does that make sense? How can we explain this better?

nwu63:
> We can also add `sens` through the `Optimization` object now. Is that worth mentioning?

I guess I didn't want to do it because it might be confusing for users. I would also like to refactor this in the future, and drop the `sens` argument when calling the optimizer. The reason they are different right now is because `objfunc` is attached to the `optProb` while `sens` is attached to the `Optimizer` instance, which then constructs an appropriate callback function via the `Gradient` class if the user wants FD or something.

doc/guide.rst
sseraj: I see what you mean about the sub-blocks. Would referring to DV groups be helpful or more confusing?

sseraj:
> I guess I didn't want to do it because it might be confusing for users.

That makes sense. I agree that it could be confusing.

doc/guide.rst
Xiaosong2105: Yes, fixed.

doc/guide.rst
Xiaosong2105: > I don't think 'amount' is the right word. It should just be 'number'.

Yes, fixed.

nwu63: > I see what you mean about the sub-blocks. Would referring to DV groups be helpful or more confusing?

Oh I like that, I will update.

nwu63: Actually, what I said is incorrect, I believe the sub-blocks themselves can also be sparse by using a pyOptSparse or SciPy sparse array. I do think this sparsity stuff should probably be refactored into the advanced section though.

sseraj: Okay, that makes sense.

sseraj (approved):

marcomangano:
Is the doc build fine and just failed to return on the UI here?

nwu63:
> Is the doc build fine and just failed to return on the UI here?

It worked fine, see log [here](https://readthedocs.com/projects/mdolab-pyoptsparse/builds/615513/).

marcomangano (approved):

marcomangano:
@nwu63 without the green check on docs I cannot merge, you need to override that

nwu63 (approved):

Issue #238, opened 2021-04-25
Duplicate entries in IPOPT history files


A-Gray-94:
## Description
History files saved when using IPOPT appear to contain duplicate function and gradient evaluation entries

### Steps to reproduce issue
_Please provide a minimum working example (MWE) if possible_

1. Run `test_hs015.py` in test directory
2. Check `hs015_IPOPT.out`, IPOPT claims to have only performed 12 function evaluations and 11 gradient evaluations
3. Open the history file in python and inspect the `db` dictionary

### Current behavior
There are 47 stored function and gradient evaluations, that's 2x what IPOPT claims + 1

![Screenshot from 2021-04-24 22-11-12](https://user-images.githubusercontent.com/20371123/115977913-21e8cd00-a54a-11eb-9aea-f44eae611369.png)

### Expected behavior
There should be 12 function evaluation entries and 11 gradient evaluation entries

My suspicion is that IPOPT makes separate calls for evaluating the objective and the constraints at the same design point and pyoptsparse is currently writing a separate history entry for each one when it should be combining them. Same for the objective and constraint gradient evaluations

### Code versions
2.6.0

nwu63:
This is a known problem and a duplicate of #182, so I'm going to close this and redirect discussions over there.

Issue #182, opened 2020-10-27
IPOPT history recording


kanekosh:
## Description
Objective and constraint calls are recorded duplicately.
When the optimizer evaluates the objective and constraint at a point `x`, it should be recorded as a single iteration. But it seems that it is recorded as two iterations with the same variable/function values.

### Steps to reproduce issue
Run any constrained problem, then check the number of function calls printed in `IPOPT.out` as well as the iterations in the OptView plot.

### Current behavior
For example, when `IPOPT.out` says it finished after 100 objective evaluations and 100 constraint evaluations, we see 200 iterations in the OptView plot as a result of the duplication.

### Expected behavior
We should have just 100 iterations in the plot.

A-Gray-94:
@kanekosh , I just discovered this issue for myself, did you ever figure out a possible fix?

kanekosh:
This bug occurs because [the IPOPT wrapper](https://github.com/mdolab/pyoptsparse/blob/master/pyoptsparse/pyIPOPT/pyIPOPT.py#L201) defines callback functions for objective(`eval_f`) and constraints (`eval_g`) separately, unlike other optimizers which define one function that returns both. Each callback wraps [_masterFanc](https://github.com/mdolab/pyoptsparse/blob/d1aea3688589815ab05b5b4b3686b093957850c0/pyoptsparse/pyOpt_optimizer.py#L198) in which we record the history. So at each point `x`, IPOPT calls `_masterFuncs` twice (once for objective, once for constraints) and duplicates the log.

We may want to unify the two callbacks into one (such as `eval_f_g`) as other optimizers do. But it needs a (major) update of the C codes.

Another way is to add the `writeHist = True/False` argument to `_masterFunc`, and enable it only from the `eval_f`. Then we don't record the history from the `eval_g` and can avoid duplication. This is just at python level and easy to do, but I'm not sure if it's a good fix.

A-Gray-94:
I don't think it's a good idea to rely on modifying how IPOPT makes evaluation calls as then we'd essentially be maintaining our own version of IPOPT no? Seems much easier to just handle it in the python layer

kanekosh:
Right we shouldn't be touching the IPOPT codes. I'm just not sure if we need to modify IPOPT itself or just our C wrappers (like pyipoptcoremodule.c).

kanekosh:
[...]

PR #239, opened 2021-04-26
Fixed IPOPT history recording

kanekosh:
## Purpose
Addresses #182 . This fixes the bug that IPOPT history files constraint duplicated entries in a _hacky_ way.

### Note
It basically checks if the current call is already recorded or not based on `x` values. If the current `x` is identical to the previous `x` at which we wrote the history, then we don't write the history.
I first wanted to check that based on the type of the call (i.e. constraints or objective), then write the history only when it is an objective (or its gradient) call. This worked fine for cold starts and the code would've been much cleaner, but it is problematic when we parse the history for hot starts.
When we don't record the constraint calls, during the hot start, we need to get info from an adjacent call counter (which is the objective call we recorded at the same `x`). But the problem is that we don't know which adjacent call counter (+1 or -1) we should refer, because we don't know which function (objective or constraints) IPOPT calls first at each iteration. It sometimes calls the objective first, but the other times it calls the constraints first. That's why we have to rely on`x` values.

Since the code is hacky, it might be better to keep the history file unfixed, but delete the duplicated entries in OptView or somewhere. What do you think @nwu63 @A-Gray-94 ?

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
This is the history file of `test_hs015.py`. It matches what IPOPT claims (12 function and 11 jacobian evaluations). The previous behavior can be found in #238 .
Screenshot from 2021-04-25 23-20-56

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimizer.py
+++ b/pyoptsparse/pyOpt_optimizer.py
@@ -90,6 +90,10 @@ class Optimizer(BaseSolver):
         # Initialize metadata
         self.metadata: Dict[str, Any] = {}
 
+        # We need to Store x at previous iteration (only matters for IPOPT)
+        self._x_prev_for_funcs = None
+        self._x_prev_for_funcsSens = None
+
     def _clearTimings(self):
         """Clear timings and call counters"""
         self.userObjTime = 0.0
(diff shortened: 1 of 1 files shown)


kanekosh:
Alternative fix in #252

Issue #191, opened 2020-12-05
Make wheels available on PyPI


goerz:
It seems that this package is a prerequisite for testing `dymos` (https://github.com/OpenMDAO/dymos/issues/471). Thus, it would be great if `pyoptsparse` was available on PyPI as well.

The installations instructions for `pyoptsparse` (`pip install .`) did not work out of the box for me, apparently due to compilation problems of the Fortran code:

[error.log](https://github.com/mdolab/pyoptsparse/files/5647217/error.log)

For packages with compiled code, I would strongly recommend uploading precompiled wheels to PyPI to make installation easy on systems that don't have the necessary compilers installed (Windows especially). I've had good experiences with using https://github.com/joerick/cibuildwheel running on Github actions to generate the wheels.

Releasing a conda-package on conda-forge may also be something to look into (but personally, I think wheels have a higher priority)

dhill2522:
I'm not sure what the project priorities are, but from a user perspective I totally agree. Having pre-built wheels available would be wonderful as helping less-advanced programmers get started using this package can be a challenge.

I would be willing to help put some work towards this as well if that would be a useful place to contribute.

nwu63:
This is something we're aware of, but unfortunately none of the maintainers have time to work on this. If anyone is willing to do this and submit a PR (maybe @dhill2522?) then that would be greatly appreciated.

I don't really know too much about how Python wheels work, or conda packages for that matter. More specifically, I'm not sure how binaries would work if the user wants to also install optional optimizers, which must be installed separately and linked via dynamic libraries.

dhill2522:
Ok. We have had some experience with PyPi and binaries with project [GEKKO](https://github.com/BYU-PRISM/GEKKO) and I think we should be able to figure something out. I will probably be relatively busy for the next week, but after that I should have a good chance to take a look at it. Certainly no guarantees I will have the time, but we will see.

dhill2522:
As a note to myself and anyone working on this. If we were to release this as a conda package rather than a PyPi package then we might be able to take advantage of solver installations already published as conda packages. For example IPOPT is available relatively easily on Mac and Linux through [cyipopt](https://github.com/matthias-k/cyipopt) (Some tweaking likely required). I realize we would probably expect people to install the solvers themselves, but it seems that pyoptsparse has some pretty inherently system-level dependencies rather than just python ones. In my experience Conda has generally worked better for that.

Are there any strong feelings on a conda package vs a PyPi one?

[...]

PR #261, opened 2021-06-06
Add conda recipe

whophil:
## Purpose
Add recipe to build pyoptsparse as a conda package.

The conda package supports all the standard (free) optimizers, plus IPOPT (which it introduces as a build and run-time [conda dependency](https://anaconda.org/conda-forge/ipopt)). This all happens under the hood, so users don't need to mess around building IPOPT from source. The particular flavor of metis/mumps dependencies you get is chosen by the maintainers of the [IPOPT conda recipe](https://github.com/conda-forge/ipopt-feedstock), so if you need full control then you'll still want to build from source.

There was some discussion about Python wheels (#191). While I think a PyPi wheel would be still be a useful thing to have, complex dependencies involving non-Python projects quickly become untenable with wheels. It would be hard to put IPOPT in a wheel without some complicated CI work; with conda, it all comes together easily.

There is one hack in this recipe, and that is related to the dependence on the `mdolab-baseclasses` package. The ideal solution would be to have `mdolab-baseclasses` available as a conda package, and to specify that as a run dependency of this package. However, that conda package doesn't exist yet. In the recipe I have embedded the source of `mdolab-baseclasses`, pinned to the `1.5.2` tag. This can be fixed if/when it is available as a conda package.

**EDIT:**
It probably makes sense to just submit this to https://github.com/conda-forge/staged-recipes, so consider this PR a demonstration. Somebody affiliated with mdolab should probably be the maintainer of that recipe. I can provide some support if needed (I would be able to open the PRs on staged-recipes for both mdolab-baseclasses and this project).

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [x] Other (packaging)

## Testing
In order to build the package, run this from the top level of the repo.

```
conda build conda.recipe
```

You will need conda and conda-build installed. This will build the package and run tests.

To install the conda package, into the active conda environment:
```
conda install -c local pyoptsparse
```

## Checklist

### DIFF
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 cd ${SRC_DIR}
-${PYTHON} -m pip install . -vv
+IPOPT_DIR=${CONDA_PREFIX} ${PYTHON} -m pip install . -vv
 
 cd ${SRC_DIR}/baseclasses
 ${PYTHON} -m pip install . -vv
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd ${SRC_DIR}
+${PYTHON} -m pip install . -vv
+
+cd ${SRC_DIR}/baseclasses
+${PYTHON} -m pip install . -vv
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -5,6 +5,8 @@ package:
 
 source:
   - path: ../
+  - patches:
+    - patch-IPOPT-setup-for-conda.patch
 
   # Embed mdolab-baseclasses package.
   # This is sub-optimal, baseclasses should be provided as a conda package as
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -20,6 +20,7 @@ requirements:
     - python
     - numpy
     - swig
+    - ipopt
     - {{ compiler('c')}}
     - {{ compiler('cxx')}}
     - {{ compiler('fortran') }}
@@ -31,6 +32,7 @@ requirements:
   run:
     - python
     - numpy
+    - ipopt
     - sqlitedict >=1.6
     - scipy >1.2
 
@@ -46,4 +48,4 @@ test:
     - parameterized
 
   commands:
-    - testflo .
+    - testflo --verbose .
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -1,0 +1,49 @@
+package:
+  name: pyoptsparse 
+  version: "{{ load_setup_py_data().version }}"
+
+
+source:
+  - path: ../
+
+  # Embed mdolab-baseclasses package.
+  # This is sub-optimal, baseclasses should be provided as a conda package as
+  # well. This may cause problems if the pyoptsparse is installed into an
+  # environment which requires other, incompatible versions of baseclasses.
+  - git_url: https://github.com/mdolab/baseclasses.git
+    git_rev: v1.5.2
+    folder: baseclasses
+
+
+requirements:
+  build:
+    - python
+    - numpy
+    - swig
+    - {{ compiler('c')}}
+    - {{ compiler('cxx')}}
+    - {{ compiler('fortran') }}
+
+  host:
+    - python
+    - numpy
+
+  run:
+    - python
+    - numpy
+    - sqlitedict >=1.6
+    - scipy >1.2
+
+test:
+  imports:
+    - pyoptsparse
+
+  source_files:
+    - test
+
+  requires:
+    - testflo
+    - parameterized
+
+  commands:
+    - testflo .
--- a/conda.recipe/patch-IPOPT-setup-for-conda.patch
+++ b/conda.recipe/patch-IPOPT-setup-for-conda.patch
@@ -1,0 +1,13 @@
+diff --git a/pyoptsparse/pyIPOPT/setup.py b/pyoptsparse/pyIPOPT/setup.py
+index f875ca2..c2ebf0d 100644
+--- a/pyoptsparse/pyIPOPT/setup.py
++++ b/pyoptsparse/pyIPOPT/setup.py
+@@ -51,7 +51,7 @@ def configuration(parent_package="", top_path=None):
+             "pyipoptcore",
+             FILES,
+             library_dirs=[IPOPT_LIB],
+-            libraries=["ipopt", "coinmumps", "coinmetis", "dl", "m", "blas", "lapack"],
++            libraries=["ipopt", "mumps_common_seq", "scotchmetis", "dl", "m", "blas", "lapack"],
+             extra_link_args=["-Wl,-rpath,%s -L%s" % (IPOPT_LIB, IPOPT_LIB)],
+             include_dirs=[numpy_include, IPOPT_INC],
+         )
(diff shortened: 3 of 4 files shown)


whophil:
This shouldn't affect coverage, there is an old test failure that appears to be stuck to this PR because I rebased a few things.

nwu63:
Hi, thanks for doing this! First, in terms of the conda recipe itself I have two comments:
1. As you pointed out, `mdolab-baseclasses` should definitely be its own conda package and that should be easy since it's pure Python
2. I would prefer a different way to configure `setup.py` to link to different libraries, but a patch is perhaps okay for now if we don't want to change the existing code. In general, I think what this illustrates is that we want a more flexible configuration option for IPOPT, such that the user is able to select the appropriate flags. Maybe this can be done via environment variables or something during building, which I think `conda` has the ability to set. I'll open an issue to track this but it doesn't really affect this particular work. This recipe also makes IPOPT a required dependency rather than optional, but maybe that's fine in the `conda` ecosystem.

Moreover, I've had some discussions with lab members regarding `conda` packages. Like you said, this would make the most sense to go into `conda-forge`, but the lab doesn't really have the resources to maintain conda packages at the moment. Would this be something that you're willing to do perhaps? If we can't find someone who's willing to maintain the conda package, then what we might do is accept this PR into another branch, and keep it there just so your work doesn't get lost, but the recipe itself won't be maintained or packaged.

whophil:
It is possible to compile the IPOPT interfaces at build time, but not require the `ipopt` package during run-time, while constraining `ipopt` to a version _if_ installed. (see https://conda-forge.org/docs/maintainer/adding_pkgs.html#constraining-packages-at-runtime). This might be the best solution here.

Regarding IPOPT's `setup.py`, I would prefer the patch solution for the `2.6.2` conda package so we can avoid the version bump churn. If future versions of `setup.py` fix this, we can certainly modify the recipe.

I am happy to submit the `mdolab-baseclasses` and `pyoptsparse` packages to conda-forge and to be listed as a maintainer. However, I can't commit to long term support and I'm really only a light user of `pyoptsparse`. So I'd feel better if there is at least _one_ additional name from your group whose name I could add to the maintainers list.

nwu63:
> It is possible to compile the IPOPT interfaces at build time, but not require the `ipopt` package during run-time, while constraining `ipopt` to a version _if_ installed. (see [conda-forge.org/docs/maintainer/adding_pkgs.html#constraining-packages-at-runtime](https://conda-forge.org/docs/maintainer/adding_pkgs.html#constraining-packages-at-runtime)). This might be the best solution here.

This would be pretty cool. Since I'm very unfamiliar with conda, I have a few more questions (assuming we get this on conda-forge):
- conda-forge will automatically build binaries linking to the conda-provided IPOPT package, right? The user just has to download the conda package and everything should work?
- What about stuff like mac OS and Windows? will those be handled automatically in some way? Or maybe we can limit ourselves to just building on Linux for now.
- I think the PR as it stands will need modifications if we are to submit to conda-forge? For example instead of specifying the source code relative to the YAML file, we have to specify some tarball from GitHub or something I guess. We can iron out the details if we wish to pursue this.

> Regarding IPOPT's `setup.py`, I would prefer the patch solution for the `2.6.2` conda package so we can avoid the version bump churn. If future versions of `setup.py` fix this, we can certainly modify the recipe.

Yeah, I agree. This is a good solution for now, but I can try to add this feature to pyOptSparse in the future.

> I am happy to submit the `mdolab-baseclasses` and `pyoptsparse` packages to conda-forge and to be listed as a maintainer. However, I can't commit to long term support and I'm really only a light user of `pyoptsparse`. So I'd feel better if there is at least _one_ additional name from your group whose name I could add to the maintainers list.

Yeah that's totally reasonable. I think I can commit to this, and in the future hopefully others in the lab can be added too. I'm just hesitant because I (and others in the lab) have very limited experience with conda, but if this is mostly just maintaining some YAML files and the build/test process is largely automated, then I'm less worried.

whophil:
I have experience with conda-packaging for internal deployment, but this would be my first time submitting to conda-forge. So prefix everything below with "AFAIK."

> * conda-forge will automatically build binaries linking to the conda-provided IPOPT package, right? The user just has to download the conda package and everything should work?

Yes, conda-forge will build the necessary interface binaries and bundle those (along with the python sources) into the conda package. When pyIPOPT is imported, it will try to import `pyipoptcore` which will look for the IPOPT library in the conda environment's `lib` folder. Conda makes sure that the library search paths are all OK.

> * What about stuff like mac OS and Windows? will those be handled automatically in some way? Or maybe we can limit ourselves to just building on Linux for now.

conda-forge supports multi-platform builds on a matrix of platforms which is defined along with the recipe. So we can have it build on whichever platforms this package supports. So there will be multiple versions of `pyoptsparse-2.6.2` available on conda-forge corresponding to different platforms. The platform is just one of the matrix options - other common matrix options include BLAS implementation (openblas vs mkl)

> * I think the PR as it stands will need modifications if we are to submit to conda-forge? For example instead of specifying the source code relative to the YAML file, we have to specify some tarball from GitHub or something I guess. We can iron out the details if we wish to pursue this.

Yep, we'll need to make some minor changes. And `mdolab-baseclasses` will have to be added and released first, as it is a dependency of this project.

> I'm just hesitant because I (and others in the lab) have very limited experience with conda, but if this is mostly just maintaining some YAML files and the build/test process is largely automated, then I'm less worried.

There shouldn't be too much babysitting or CI administration, as that's all handled by conda-forge. Hopefully it isn't too much work, and I will support the best that I can.

I'll put in the PR for `mdolab-baseclasses` today, and `pyoptsparse` whenever that one is merged in. I will list you as a maintainer on both and ping you on the PRs. Or would you prefer I use @pyoptsparse_maintainers?.

I'm happy to discuss any questions you have about those PRs, or conda packaging in general. I think it would be a huge benefit to the community if more of the mdolab tools were available on conda-forge, and this is probably a good first step.

I appreciate all the effort of you and your group. So thank you, and thanks for entertaining this!

nwu63:
> There shouldn't be too much babysitting or CI administration, as that's all handled by conda-forge. Hopefully it isn't too much work, and I will support the best that I can.
> 
> I'll put in the PR for `mdolab-baseclasses` today, and `pyoptsparse` whenever that one is merged in. I will list you as a maintainer on both and ping you on the PRs. Or would you prefer I use @pyoptsparse_maintainers?.
> 
> I'm happy to discuss any questions you have about those PRs, or conda packaging in general. I think it would be a huge benefit to the community if more of the mdolab tools were available on conda-forge, and this is probably a good first step.
> 
> I appreciate all the effort of you and your group. So thank you, and thanks for entertaining this!

Sounds like a plan! I don't think the maintainer group will work since that's specific to the `mdolab` organization on GitHub. Please put `nwu63` and `marcomangano` as maintainers please.

No thank you for getting this started, it has been requested several times but we were sort of waiting for a PR since we were unfamiliar with conda ourselves. We can discuss the specifics on the PR to `staged-recipes`, but I imagine there would be issues with how we test our codes for example (ideally we force IPOPT tests under conda, which are optional right now). None of these are blockers however and we can get something that works first before we cover everything.

As for other mdolab codes, we had talked about that internally, but since it doesn't usually fall under our use case within the lab there has been low enthusiasm to do that. But I think pyOptSparse occupies a unique place since 1) it's used by many outside the lab, and 2) it would benefit a lot from conda. Let's start here and see where things go, it's possible that in the future we may support more packages.

whophil:
The conda-forge recipe is now open here: https://github.com/conda-forge/staged-recipes/pull/15239. Closing this PR!

PR #265, opened 2021-06-21
Remove other linked libraries when building `ipoptcore`

### LINKED ISSUE #262: More flexible configurations for building IPOPT
# Description of feature
Right now, we hard-code the libraries when building IPOPT: `libraries=["ipopt", "coinmumps", "coinmetis", "dl", "m", "blas", "lapack"]` but the user may wish to use other linear solvers for example.

# Potential solution
We could rely on certain environment variables that are set at build time to change these flags without having to modify `setup.py`. I don't have a good idea on what the naming scheme should be, though.

nwu63:
## Purpose
Closes #262.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Let's see if this builds fine. If IPOPT tests all pass then I think we are good to go.

## Checklist

### DIFF
--- a/pyoptsparse/pyIPOPT/setup.py
+++ b/pyoptsparse/pyIPOPT/setup.py
@@ -51,7 +51,7 @@ def configuration(parent_package="", top_path=None):
             "pyipoptcore",
             FILES,
             library_dirs=[IPOPT_LIB],
-            libraries=["ipopt", "coinmumps", "coinmetis", "dl", "m", "blas", "lapack"],
+            libraries=["ipopt"],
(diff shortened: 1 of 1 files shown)


marcomangano (approved):
It works so I guess we are good, but can you clarify what is going on? Are we checking/building these (or similar) libraries directly when building IPOPT?
Basically `pyIPOPT` is only looking for `ipopt` lib, and everything else is handled at a lower level, without any implication for the python layer?

nwu63:
Yeah the idea is that we should only need to link to `ipopt`, libraries such as `mumps` should be linked to from `ipop` so we don't have to add those linker flags here.

ArshSaja (approved):

PR #266, opened 2021-06-24
Allow objFun pickling if possible 

### LINKED ISSUE #264: Why skip objFun when serializing Optimization?
I have a `Optimization` with a very cheap `objFun` and am using `concurent.futures` to solve multi-starts in parallel. I've noticed that the `Optimization` arrives on the executor side without its `objFun`.

This is due to the following:
https://github.com/mdolab/pyoptsparse/blob/master/pyoptsparse/pyOpt_optimization.py#L1785

Is there a particular reason the `objFun` is skipped in the serialization of `Optimization`? Perhaps not every function is picklable, but some are.

Proposal: Test if `objFun` can be pickled; put it into the state dict if it can.

sseraj:
## Purpose
Closes #264 by allowing objFun to be pickled if possible. We use a blanket Exception to avoid issues with the tests and MACH.

## Type of change
What types of change is it?

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimization.py
+++ b/pyoptsparse/pyOpt_optimization.py
@@ -2,9 +2,9 @@
 from collections import OrderedDict
 import copy
 import os
+import pickle
 from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
 import warnings
-import pickle
 
 # External modules
 import numpy as np
--- a/pyoptsparse/pyOpt_optimization.py
+++ b/pyoptsparse/pyOpt_optimization.py
@@ -1786,7 +1786,10 @@ class Optimization(object):
         keysToRemove = ["comm"]
         try:
             pickle.dumps(self.objFun)
-        except pickle.PicklingError:
+        except Exception:
+            # Use a blanket exception because pickle errors are unreliable
+            # Tests raise RecursionError
+            # mpi4py raises TypeError
             keysToRemove.append("objFun")
         for key in keysToRemove:
             if key in d.keys():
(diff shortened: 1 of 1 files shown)


nwu63 (approved):

marcomangano (approved):

pyoptsparse/pyOpt_optimization.py
```diff
@@ -1782,7 +1783,15 @@ def __getstate__(self) -> dict:
         The un-serializable fields are deleted first.
         """
         d = copy.copy(self.__dict__)
-        for key in ["comm", "objFun"]:
+        keysToRemove = ["comm"]
+        try:
+            pickle.dumps(self.objFun)
```
whophil, line 1788: Seems like this will result in `objFun` being pickled twice (if it is serializable).

How about the following
```python
try:
    d['objFun'] = pickle.dumps(self.objFun)
except Exception:
    keysToRemove.append("objFun")
```

with

```python
def __setstate__(self, state):
    self.__dict__ = state
    # at this point, objFun is still a string; deserialize
    self.objFun = dill.loads(self.objFun)
```
I would guess there is only a minor performance difference.

sseraj: It's true that `objFun` will be pickled twice, but I'm fine with this to keep the code simpler.

nwu63: I think it's going to get pickled twice anyways, since the whole dictionary that is returned by `__getstate__` will be pickled anyhow. Unless there are noticeable performance issues I think the current code is probably okay.

PR #271, opened 2021-08-28
Convert to f-strings

nwu63:
## Purpose
I used `flynt` to convert old-style strings to f-strings. In addition, there were a bunch of manual fixes, so it would be helpful of reviewers are careful in reviewing this to make sure I didn't break anything.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
N/A

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_variable.py
+++ b/pyoptsparse/pyOpt_variable.py
@@ -82,7 +82,7 @@ class Variable:
         if self.type == "d":
             res += "	 "
             res += str(self.name).center(15)
-            res += f"{self.type:>25}{self.value:20f} {min(self.choices):14.2e} {max(self.choices):12.2e}\n"
+            res += f"{self.type:>25}{self.choices[int(self.value)]:20f} {min(self.choices):14.2e} {max(self.choices):12.2e}\n"
         else:
             lower = self.lower
             upper = self.upper
--- a/pyoptsparse/pyOpt_variable.py
+++ b/pyoptsparse/pyOpt_variable.py
@@ -82,12 +82,7 @@ class Variable:
         if self.type == "d":
             res += "	 "
             res += str(self.name).center(15)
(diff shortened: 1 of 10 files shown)


pyoptsparse/pyOpt_variable.py
```diff
@@ -82,12 +82,7 @@ def __str__(self) -> str:
         if self.type == "d":
             res += "	 "
             res += str(self.name).center(15)
-            res += "{:>25}{:20f} {:14.2e} {:12.2e} \n".format(
```
marcomangano, line 85: why did you change `self.choices[int(self.value)]` into just `self.value`?

marcomangano (commented):
I checked all the changes and they look sound to me, can you confirm that the change below was intentional?

pyoptsparse/pyOpt_variable.py
nwu63: Good question. It seems like `self.value` is already an integer from above, but at the same time I'm also somewhat certain this stuff is broken since `value` and `self.value` are not the same (in the `__init__`). We do not have any tests for integer DVs I think.

marcomangano: Yeah but I think in the first case `self.choices` was a list, and now it's replaced by an integer? Anyway, if we don't have tests and we don't have working examples with this bit of code this is very hard to debug

nwu63: Yeah you know what I'll just put this back so this PR is not making any syntax changes. I know we should have tests for this but it's very low on my priority list.

marcomangano: yeah I am on the same page, this is not urgent now

marcomangano (approved):

sseraj (approved):

PR #275, opened 2021-11-15
remove deepcopy of sensitivity dict to improve performance

Kenneth-T-Moore:
## Purpose

While doing some profiling, I discovered that the deepcopy of the sensitivity dictionary during `_masterFunc2` is a rather slow operation. On my problem, an 11.5 second optimization was spending 1.95 seconds in that deepcopy. Removing it and replacing with just a reference dropped my runtime to 9.5 seconds.  Obviously this is peanuts if you are optimizing a problem with long objective and gradient compute times, but if you are performing many optimizations on small fast problems, that extra couple of seconds per run is important.

I think replacing with a reference is safe in most cases, if not all. The user-provided (or openmdao-provided) gradient function assembles that dictionary, and I can't think of any reasonable cases where something is going to change those values before the next gradient request.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

This is just a performance change, so I verified that existing tests pass.

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimizer.py
+++ b/pyoptsparse/pyOpt_optimizer.py
@@ -468,7 +468,9 @@ class Optimizer(BaseSolver):
                 self.userSensCalls += 1
 
                 # User values are stored immediately
-                self.cache["funcsSens"] = copy.deepcopy(funcsSens)
+                # deepcopy of the sens dictionary is slow, so just reference it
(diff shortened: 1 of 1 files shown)


nwu63 (approved):
Thanks for this, LGTM

sseraj (approved):

PR #276, opened 2021-11-27
Fixed  an issue for the save_tec func in OptView

friedenhe:
Choosing the save tec file button will not work for OptView.py. It returned a "dict_keys object is not subscriptable" error. Need to convert the dict_keys to a list.

## Purpose
Fixed a bug for the save_tec function in OptView

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pyoptsparse/postprocessing/OptView.py
+++ b/pyoptsparse/postprocessing/OptView.py
@@ -747,7 +747,7 @@ class Display(OVBaseClass):
             for name in values:
                 dat[name] = self.var_data[name]
 
-        keys = dat.keys()
+        keys = list(dat.keys())
 
         num_vars = len(keys)
         num_iters = len(dat[keys[0]])


sseraj (approved):

pyoptsparse/postprocessing/OptView.py
```diff
@@ -747,7 +747,7 @@ def save_tec(self):
             for name in values:
                 dat[name] = self.var_data[name]
 
-        keys = dat.keys()
+        keys = list(dat.keys())
```
nwu63, line 750: Given that this is unsorted, are we sure that for example line 753 is correct?

friedenhe: This should be fine. Although dat.keys() are not sorted, all its arrays have the same size. I have tested it with more than one key and it worked fine.

nwu63: Ah okay gotcha.

nwu63 (approved):

pyoptsparse/postprocessing/OptView.py
friedenhe: BTW: I think the save tec file is a very useful function for OptView. However, it is not in the OptView_dash version. I can not save the optimality to files for plotting. Maybe save_tec_file is a new feature worth migrating to OptView_dash

nwu63: Yeah OptView in general is a bit of a mess. We should clean that up, modernize it and make sure everything's consistent, but we just haven't had the chance to.

PR #277, opened 2021-12-20
Add message attribute to Exceptions

Laurentww:
## Purpose

Adding a message attribute to PyOptSparse Exceptions which allows for parsing of exceptions.

The motivation behind adding the message attribute is to be able to differentiate between `Error`s during testing. For example, to know whether an error is due to the SNOPT libraries not being available or due to problem option incompatibility.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Raise `Error()` during the evaluation of the problem objective or sensitivities and check that the raised error instance has a `message` attribute.

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_error.py
+++ b/pyoptsparse/pyOpt_error.py
@@ -23,7 +23,11 @@ class Error(Exception):
                 i += len(word) + 1
         msg += " " * (79 - i) + "|\n" + "+" + "-" * 78 + "+" + "\n"
         print(msg)
-        Exception.__init__(self)
+        self.message = msg
+        super(Error, self).__init__(message)
+
+    def __str__(self):
+        return self.message
 
 
 class pyOptSparseWarning:
--- a/pyoptsparse/pyOpt_error.py
+++ b/pyoptsparse/pyOpt_error.py
@@ -24,7 +24,7 @@ class Error(Exception):
         msg += " " * (79 - i) + "|\n" + "+" + "-" * 78 + "+" + "\n"
         print(msg)
         self.message = msg
-        super(Error, self).__init__(message)
+        super().__init__(message)
 
     def __str__(self):
         return self.message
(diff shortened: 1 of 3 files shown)


pyoptsparse/pyOpt_error.py
```diff
@@ -23,7 +23,11 @@ def __init__(self, message):
                 i += len(word) + 1
         msg += " " * (79 - i) + "|\n" + "+" + "-" * 78 + "+" + "\n"
         print(msg)
-        Exception.__init__(self)
+        self.message = msg
```
nwu63, line 26: Do you want to store the formatted message, or the raw `message` string as the attribute? The latter may be easier to work with if you want to parse it.

pyoptsparse/pyOpt_error.py
```diff
@@ -23,7 +23,11 @@ def __init__(self, message):
...
+        super(Error, self).__init__(message)
```
nwu63, line 27: I would prefer the more modern `super().__init__(message)` here

nwu63 (commented):
See comments below. Can you talk about the motivation behind this change, and how you would go about using this?

pyoptsparse/pyOpt_error.py
Laurentww: That makes sense, thanks

pyoptsparse/pyOpt_error.py
Laurentww: Left it in there for clarity, but I agree

Laurentww:
Updated the PR purpose section

nwu63:
Ah, I see the motivation now. This is something that we had discussed internally also---using just the same `Error` class for everything is probably a bad idea. In the long term, I would actually like to go back to built-in exceptions for most stuff, and maybe even create a few specific exception classes for specific issues (such as error importing the compiled libraries). That way, the error message/catching can be done more programmatically than parsing the message and relying on the message being a certain way.

That being said, I'm fine with this as is, but do watch out for release messages in the future if we decide to overhaul the error system. And if you ever feel like working on that feel free to reach out, we're always happy to take contributions :)

nwu63 (approved):
Looks good, thanks for adding this to the tests too

marcomangano (approved):

PR #279, opened 2022-01-28
Suppress SNOPT minor print log

nwu63:
## Purpose
I changed the default SNOPT options to suppress the minor print log

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Try it locally, the minor printing log no longer appears.

## Checklist

### DIFF
--- a/doc/optimizers/SNOPT_options.yaml
+++ b/doc/optimizers/SNOPT_options.yaml
@@ -7,8 +7,11 @@ iSumm:
 Print file:
   desc: Print file name
 
-Summary file: 
+Summary file:
   desc: Summary file name
+
+Minor print level:
+  desc: Minor iterations print level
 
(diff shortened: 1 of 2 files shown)


sseraj (approved):
The minor iteration log usually clutters the output without adding much value so I support this change.

marcomangano (approved):
Agreed, this is the best default behavior. Still not sure why the minor iters were just appearing "randomly" though.

nwu63:
> Agreed, this is the best default behavior. Still not sure why the minor iters were just appearing "randomly" though.

They were not! By default, I think the log is printed every 100 minor iterations. There are some major iterations (especially later ones) where the number of minor iterations per major iteration is < 100, so nothing is shown.

Issue #220, opened 2021-03-20
Refactor optimizer wrappers


nwu63:
# Description of feature
There is a ton of shared code between different optimizer wrappers:
- setting up the parameters
- comm splitting and bcast after
- creating the solution after

There are also tons of class attributes which are set by each optimizer without using the class structure properly. All these can be reorganized with significantly less code duplication.

# Potential solution
- Create a `__call__` function in the parent class, which does the shared stuff:
	- This can call optimizer-specific functions which does the actual work (generating inputs, calling the optimizer, etc.)
- In the init, properly set the class attributes.

PR #281, opened 2022-02-04
Added iteration timing to history file

### LINKED ISSUE #147: Add iteration timing
## Type of issue
What types of issue is it?

## Description
I propose that we add a new key called `time` under each `callCounter`, which stores the wall time elapsed in seconds since the beginning of the optimization. This would allow an optimization history to be plotted against wall time, in addition to the callCounter available.

nwu63:
## Purpose
This was a popular feature request. Closes #147. The history file will now have `time` embedded in each iteration, which is helpful for plotting values against time instead of just iterations. Unfortunately the implementation is a bit crude since the same boilerplate code is in each optimizer wrapper---see #220. Nevertheless, this works and I've added some tests also.

I also added a basic`tearDown` implementation to clean up some of the generated test files, but this is incomplete.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I added some tests.

## Checklist

### DIFF
--- a/tests/testing_utils.py
+++ b/tests/testing_utils.py
@@ -1,6 +1,6 @@
 # Standard Python modules
-import unittest
 import os
+import unittest
 
 # External modules
 import numpy as np
--- a/tests/testing_utils.py
+++ b/tests/testing_utils.py
@@ -383,6 +383,13 @@ class OptTest(unittest.TestCase):
         # getValues checks
         val = hist.getValues()
 
+        # timing checks
+        times = hist.getValues(names="time", major=False)["time"]
+        # the times should be monotonically increasing
+        self.assertTrue(np.all(np.diff(times) > 0))
+        # the final time should be close to the metadata time
+        assert_allclose(times[-1], metadata["optTime"], rtol=1e-3, atol=1e-1)
+
         # this check is only used for optimizers that guarantee '0' and 'last' contain funcs
         if self.optName in ["SNOPT", "PSQP"]:
             val = hist.getValues(callCounters=["0", "last"], stack=True)
--- a/tests/testing_utils.py
+++ b/tests/testing_utils.py
@@ -1,5 +1,6 @@
 # Standard Python modules
 import unittest
+import os
 
 # External modules
 import numpy as np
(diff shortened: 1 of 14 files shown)


ArshSaja:
Some style checks are failing. Is it okay to move on ?

nwu63:
I've fixed all the violations

tests/testing_utils.py
```diff
@@ -432,3 +443,13 @@ def optimize_with_hotstart(self, tol, optOptions=None, x0=None):
         # we should have zero actual function/gradient evaluations
         self.assertEqual(self.nf, 0)
         self.assertEqual(self.ng, 0)
+
+    def tearDown(self):
+        # remove history file
```
marcomangano, line 448: What is missing here? It looks like it is covering all the generated files

marcomangano (commented):
This is nice! ...but, given that we already have `metadata['optTime']`, wouldn't it be more useful to store the single iteration time rather than the cumulative time at iteration n? I know it's an easy post-production step and might mess with the test you set up, but I feel the info would me easier to interpret, especially if using OptView or `dictview`

nwu63:
> This is nice! ...but, given that we already have `metadata['optTime']`, wouldn't it be more useful to store the single iteration time rather than the cumulative time at iteration n? I know it's an easy post-production step and might mess with the test you set up, but I feel the info would me easier to interpret, especially if using OptView or `dictview`

I think from a postprocessing perspective cumulative time makes more sense, as you would usually want to plot DV etc. against cumulative time to replace the iteration counters. What would be the use case for single iteration time?

tests/testing_utils.py
nwu63: Didn't get rid of all the files for me when I tried this. Don't have time to fix this though.

marcomangano:
> > This is nice! ...but, given that we already have `metadata['optTime']`, wouldn't it be more useful to store the single iteration time rather than the cumulative time at iteration n? I know it's an easy post-production step and might mess with the test you set up, but I feel the info would me easier to interpret, especially if using OptView or `dictview`
> 
> I think from a postprocessing perspective cumulative time makes more sense, as you would usually want to plot DV etc. against cumulative time to replace the iteration counters. What would be the use case for single iteration time?

Oh I see, makes sense! From my perspective, because my cases are annoyingly long, it could be useful to monitor iter-time vs iteration as well. But you are right, from a general user perspective the cumulative time might be more relevant

ArshSaja (approved):
Thanks Neil..

marcomangano (approved):

PR #282, opened 2022-02-15
Enable hotStart option for ALPSO optimization

Laurentww:
## Purpose

To enable hot-starting of ALPSO optimization.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [x] Other (please describe)

Enabling of a disabled feature

## Checklist

### DIFF
--- a/pyoptsparse/pyALPSO/pyALPSO.py
+++ b/pyoptsparse/pyALPSO/pyALPSO.py
@@ -81,7 +81,7 @@ class ALPSO(Optimizer):
         }
         return defOpts
 
-    def __call__(self, optProb, storeHistory=None, **kwargs):
+    def __call__(self, optProb, storeHistory=None, hotStart=None, **kwargs):
         """
         This is the main routine used to solve the optimization
         problem.
@@ -95,6 +95,16 @@ class ALPSO(Optimizer):
         storeHistory : str
             File name of the history file into which the history of
             this optimization will be stored
+
+        hotStart : str
+            File name of the history file to "replay" for the
+            optimization.  The optimization problem used to generate
+            the history file specified in 'hotStart' must be
+            **IDENTICAL** to the currently supplied 'optProb'. By
+            identical we mean, **EVERY SINGLE PARAMETER MUST BE
+            IDENTICAL**. As soon as he requested evaluation point
+            from ALPSO does not match the history and function
+            evaluations revert back to normal evaluations.
 
         Notes
         -----
@@ -114,7 +124,7 @@ class ALPSO(Optimizer):
         self.optProb = optProb
         self.optProb.finalize()
         # Set history/hotstart/coldstart
-        self._setHistory(storeHistory, None)
+        self._setHistory(storeHistory, hotStart)
         self._setInitialCacheValues()
 
         if len(optProb.constraints) == 0:


Laurentww:
Thanks for reviewing! 

I added a test using the simple Sphere problem because on the Rosenbrock problem it was increasingly difficult to tune the ALPSO parameters in order to pass the test, bearing in mind reasonable tolerances and computational cost. Still, I'm not sure if the added feature rectifies the added test compute-time.

Laurentww:
Thanks for reviewing!

I added a simple Sphere problem because it was difficult to tune the ALPSO parameters for the Rosenbrock problem, keeping in mind reasonable test tolerances and added computational cost. Still, I’m not sure if the test compute-time justifies the added feature.

marcomangano:
> Thanks for reviewing!
> 
> I added a simple Sphere problem because it was difficult to tune the ALPSO parameters for the Rosenbrock problem, keeping in mind reasonable test tolerances and added computational cost. Still, I’m not sure if the test compute-time justifies the added feature.

Thanks for the extra effort, that's a nice addition! I tried the changes locally and the new test only adds 5 seconds to the total time. For the Azure tests we use the `--pre-announce` option, which runs testflo on a single core. That explains the longer time needed by the new test on this PR. Anyway, the time increase in that case is still negligible in my opinion, so I feel it's worth keeping the new test.

tests/testing_utils.py
```diff
@@ -184,9 +184,10 @@ def assert_solution_allclose(self, sol, tol, partial_x=False):
         # now we assert against the closest solution
         # objective
         # sol.fStar was broken for earlier versions of SNOPT
-        if self.optName == "SNOPT" and parse_version(self.optVersion) >= parse_version("7.7.7"):
-            assert_allclose(sol.fStar, self.fStar[self.sol_index], atol=tol, rtol=tol)
-        sol_objectives = np.array([sol.objectives[key].value for key in sol.objectives])
+        if self.optName == "SNOPT" and parse_version(self.optVersion) < parse_version("7.7.7"):
```
nwu63, line 187: Good fix!

tests/testing_utils.py
```diff
@@ -412,7 +413,8 @@ def optimize_with_hotstart(self, tol, optOptions=None, x0=None):
         sol = self.optimize(storeHistory=True, optOptions=optOptions, setDV=x0)
         self.assert_solution_allclose(sol, tol)
         self.assertGreater(self.nf, 0)
-        self.assertGreater(self.ng, 0)
+        if self.optName != "ALPSO":
```
nwu63, line 416: Instead of adding these checks specifically for ALPSO, I would prefer if we defined a set of gradient-free optimizers after [this line](https://github.com/Laurentww/pyoptsparse/blob/master/tests/testing_utils.py#L129), e.g. `GRAD_FREE_OPTIMIZERS = {"NSGA2", "ALPSO"}`, then we can check with `if self.optName in GRAD_FREE_OPTIMIZERS:`

nwu63 (changes requested):
Thanks for adding the test, see minor comments below.

tests/testing_utils.py
Laurentww: That clarifies it nicely, thanks for suggesting. I added it in the last commit.

nwu63 (approved):

marcomangano (approved):

PR #284, opened 2022-03-09
Add support for snstop and hot start in pySNOPT

nwu63:
## Purpose
This PR adds two features to the SNOPT wrapper:
- allowing a user-specified SNSTOP handle to monitor or terminate the optimization
- support for hot start in SNOPT. Note that this is a feature in SNOPT and separate from the `hotStart` option offered by pyOptSparse which replays the optimization history

## Expected time until merged
By end of the week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -484,11 +521,28 @@ def __call__(
...
+        if self.getOption("Return work arrays"):
+            restartDict = {
+                "cw": cw,
+                "iw": iw,
+                "rw": rw,
+                "xs": xs,
+                "hs": hs,
+                "pi": pi,
+            }
+            return commSol, restartDict
```
marcomangano, line 543: Is having a different number of returned arguments depending on an option the best practice?
We could work around this by having another couple of methods called `setRestartDict` and `getRestartDict` to avoid this user-facing API change

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -614,7 +668,18 @@ def _snstop(self, ktcond, mjrprtlvl, minimize, n, nncon, nnobj, ns, itn, nmajor,
...
-        iabort = 0
+                # this adds funcs etc. to the iterDict by fetching it from the history file
+                iterDict = self.hist.read(callCounter)
+        snstop_handle = self.getOption("snSTOP function handle")
+        if snstop_handle is not None:
+            if not self.storeHistory:
+                raise Error("snSTOP function handle must be used with storeHistory=True")
+            iabort = snstop_handle(iterDict)
+            # if no return, assume everything went fine
+            if iabort is None:
```
marcomangano, line 679: this if statement is potentially unnecessary, `iabort` is `None` if `snstop_handle` is so, and that is already been checked within this section of the code

pyoptsparse/pySNOPT/source/f2py/snopt.pyf
```diff
@@ -396,50 +396,50 @@ python module snopt ! in
...
+            integer intent(out) :: info
+            integer intent(out) :: mincw
+            integer intent(out) :: miniw
+            integer intent(out) :: minrw
+            integer intent(out) :: ns
+            integer intent(out) :: ninf
+            double precision intent(out) :: sinf
             double precision intent(out) :: obj
-            character*8 intent(inplace), dimension(lencu) :: cu
-            integer optional,check(len(cu)==lencu),depend(cu) :: lencu=len(cu)
```
marcomangano, line 432: Are we changing the shape of this array with this PR?

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -436,22 +473,23 @@ def __call__(
...
             inform = np.array([-1], np.intc)
             mincw = np.array([0], np.intc)
             miniw = np.array([0], np.intc)
             minrw = np.array([0], np.intc)
-            nS = np.array([0], np.intc)
-            ninf = np.array([0], np.intc)
-            sinf = np.array([0.0], np.float)
 
+            if restartDict is not None:
+                hs = restartDict["hs"]
```
marcomangano, line 482: We could add an inline comment, just to specify what these SNOPT parameters are

marcomangano (changes requested):
I am still not 100% getting what is going on under the hood, but overall makes sense.
See the comments below, I haven't tested the code yet but I am looking fwd to that.

The main issue I see is that I think we should not change the output of `__call__()`.
A second concern is that we now have two separate hotstart procedures for SNOPT 1) the one you are enabling in this PR (using `"Start": "Hot"` in the options and passing the `restartDict`) and 2) the general one done in the python wrapper by replaying the optimization history (passing). We should better clarify this - or imho even disable the python-based hotstart for SNOPT?

(P.S. are we actually supporting SNOPT "proper" warm start?)

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -484,11 +521,28 @@ def __call__(
...
         # Communication solution and return
         commSol = self._communicateSolution(sol)
-
-        return commSol
+        cw = self.optProb.comm.bcast(cw, root=0)
+        iw = self.optProb.comm.bcast(iw, root=0)
+        rw = self.optProb.comm.bcast(rw, root=0)
+        xs = self.optProb.comm.bcast(xs, root=0)
+        hs = self.optProb.comm.bcast(hs, root=0)
+        pi = self.optProb.comm.bcast(pi, root=0)
```
sseraj, line 533: Could you just `bcast` `restartDict` instead of all of these?

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -568,10 +622,10 @@ def _getPenaltyParam(self, iw, rw):
...
-                iobj, scaleobj, objadd, fobj, fmerit, penparm, step, primalinf, dualinf, maxvi, maxvirel, hs,
-                locj, indj, jcol, scales, bl, bu, fx, fcon, gcon, gobj, ycon, pi, rc, rg, x, cu, iu, ru, cw, iw, rw):
-    # fmt: on # noqa: E115
+    def _snstop(self, ktcond, mjrprtlvl, minimize, n, nncon, nnobj, ns, itn, nmajor, nminor, nswap, condzhz, iobj, scaleobj,
+                objadd, fobj, fmerit, penparm, step, primalinf, dualinf, maxvi, maxvirel, hs, locj, indj, jcol, scales, bl, bu, fx, fcon, gcon, gobj, ycon,
+                pi, rc, rg, x, cu, iu, ru, cw, iw, rw):
+        # fmt: on
         """
         This routine is called every major iteration in SNOPT, after solving QP but before line search
         Currently we use it just to determine the correct major iteration counting,
```
sseraj, line 631: This docstring should be updated

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -436,22 +473,23 @@ def __call__(
...
             timeA = time.time()
             # fmt: off
-            obj = snopt.snkerc(start, nnCon, nnObj, nnJac, iObj, ObjAdd, ProbNm,
-                               self._userfg_wrap, snopt.snlog, snopt.snlog2, snopt.sqlog, self._snstop,
-                               Acol, indA, locA, bl, bu, Names, hs, xs, pi, rc, inform,
-                               mincw, miniw, minrw, nS, ninf, sinf, cu, iu, ru, cw, iw, rw)
+            hs, xs, pi, rc, inform, mincw, miniw, minrw, nS, ninf, sinf, obj = snopt.snkerc(
+                start, nnCon, nnObj, nnJac, iObj, ObjAdd, ProbNm,
+                self._userfg_wrap, snopt.snlog, snopt.snlog2, snopt.sqlog, self._snstop,
+                Acol, indA, locA, bl, bu, Names, hs, xs, pi, cu, iu, ru, cw, iw, rw,
```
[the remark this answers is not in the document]
nwu63: This change (and the comment you had above about removing other vars) is because, these variables are actually outputs from `snkerc`, not inputs. As a result, I have moved them from the argument side to the output side in the `snkerc` call. They are also not used, because we already call `snmemb` manually. I opted to not replace these variables with just `_` because I felt it was a bit more explicit in matching the function signature in Fortran, but I'm also okay with replacing them.

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -436,22 +473,23 @@ def __call__(
 
             Names = np.array(["        "], "c")
             pi = np.zeros(ncon, np.float)
-            rc = np.zeros(nvar + ncon, np.float)
             inform = np.array([-1], np.intc)
             mincw = np.array([0], np.intc)
             miniw = np.array([0], np.intc)
             minrw = np.array([0], np.intc)
-            nS = np.array([0], np.intc)
```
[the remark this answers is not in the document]
nwu63: Agreed, these can be removed.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: Sure, these constitute the "current state", so we pass them into SNOPT when calling `snkerc`. This is also why they are set differently from the work arrays.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: This is meant to catch the case where the `snstop_handle` is provided by the user, but only used for monitoring/printing and the function does not return anything. Since SNOPT demands a return code, we set it to 0 here.

pyoptsparse/pySNOPT/pySNOPT.py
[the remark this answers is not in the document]
nwu63: You can also see in the pyf file that the intents have been more clearly labelled.

pyoptsparse/pySNOPT/source/f2py/snopt.pyf
nwu63: Are you talking about the array `cw`? No, it is still `lencu`, but the size of each entry was previously specified via `character*8` and now via the second argument to `dimension(lencu, 8)`. I think this is the preferred, more modern approach.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: I personally think this is fine, since the user has control over this by specifying a Python option. However, I'm open to changing the API to make it more user-friendly.

Instead of having a separate function, what about specifying the file path where pySNOPT would write the dictionary as a file (either pickle or numpy format, we can decide). Then, we can have `__call__(..., restartFile=<file path>)` accept the file path. That way, saving/loading the restart file would be handled by pyOptSparse instead of the user.

nwu63:
> The main issue I see is that I think we should not change the output of `__call__()`.

I commented on this separately below.

> A second concern is that we now have two separate hotstart procedures for SNOPT 1) the one you are enabling in this PR (using `"Start": "Hot"` in the options and passing the `restartDict`) and 2) the general one done in the python wrapper by replaying the optimization history (passing). We should better clarify this - or imho even disable the python-based hotstart for SNOPT?

I don't think we should disable the python-based one, since I see uses for that as well. I think we can explain this better to make sure there's no confusion, but it's a bit unfortunate that the names are identical.

> (P.S. are we actually supporting SNOPT "proper" warm start?)

No we are not. Are you suggesting that we remove the "warm" option? I would be okay with that.

pyoptsparse/pySNOPT/source/f2py/snopt.pyf
```diff
@@ -396,50 +396,50 @@ python module snopt ! in
             double precision dimension(lenrw) :: rw
             integer optional,check(len(rw)>=lenrw),depend(rw) :: lenrw=len(rw)
         end subroutine snoptc
-        subroutine snkerc(start,m,n,ne,nname,nncon,nnobj,nnjac,iobj,objadd,prob,userfg,snlog,snlog2,sqlog,snstop,jcol,indj,locj,bl,bu,names,hs,x,pi,rc,inform,mincw,miniw,minrw,ns,ninf,sinf,obj,cu,lencu,iu,leniu,ru,lenru,cw,lencw,iw,leniw,rw,lenrw) ! in :snopt:snoptc.f
+        subroutine snkerc(start,m,n,nej,nnames,nncon,nnobju,nnjac,iobju,objuadd,prob,userfg,snlog,snlog2,sqlog,snstop,jcol,indj,locj,bl,bu,names,hs,x,pi,rc,info,mincw,miniw,minrw,ns,ninf,sinf,obj,cu,lencu,iu,leniu,ru,lenru,cw,lencw,iw,leniw,rw,lenrw) ! in :snopt7_python:snopt7_python.f90
```
sseraj, line 399: Is this supposed to duplicate the official SNOPT python wrapper? I searched for `snkerc` [here](https://github.com/snopt/snopt-python/blob/master/solvers/f2py/snopt7_python.f90) but could not find it.

tests/testing_utils.py
```diff
@@ -251,7 +251,7 @@ def update_OptOptions_output(self, optOptions):
             optOptions[optionName] = f"{self.id()}{suffix}"
         return optOptions
 
-    def optimize(self, sens=None, setDV=None, optOptions=None, storeHistory=False, hotStart=None):
+    def optimize(self, sens=None, setDV=None, optOptions=None, storeHistory=False, hotStart=False):
```
sseraj, line 254: Was this a bug in the tests previously?

pyoptsparse/pySNOPT/pySNOPT.py
sseraj: I am fine with the current return approach. Are you using file I/O for your cases @nwu63 or passing the dictionary through memory? I can see the file approach being better if the use case requires the user to do it anyway.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: Absolutely, I'll make the change. You can see how old the code here is --- this is before I really understood how MPI works.

pyoptsparse/pySNOPT/source/f2py/snopt.pyf
nwu63: Not really, they don't support `snkerc` but call `snoptc` which wraps it. We use `snkerc` because it allows us to pass in a user-specified SNSTOP function (and log functions but we don't do that).

This change is cosmetic and merely mirrors the variable names within the SNOPT source code to make things consistent.

tests/testing_utils.py
nwu63: No, we essentially check `bool(hotStart)` and in that case `None` and `False` behave the same way. I just made this more explicit and to match the docstrings.

pyoptsparse/pySNOPT/pySNOPT.py
marcomangano: My idea of adding the additional methods was just to have a "lighter" modification than the internal I/O, but if you are willing to do so I think that the `<file-path>` approach would be the smoothest approach on the user side! Not sure what data format is best, in my mind 1) we should not add more dependencies (e.g. json?) 2) (optional) it would be nice for the database to be human readable or easily accessible in case we want to double check what's in there

pyoptsparse/pySNOPT/pySNOPT.py
marcomangano: Oh, I did not realize that, nice. I think that in the longer term might be nice to have some docs/examples on `snstop_handle`, it sounds like a pretty powerful feature

pyoptsparse/pySNOPT/source/f2py/snopt.pyf
sseraj: I see. What does this comment mean though `! in :snopt7_python:snopt7_python.f90` ?

marcomangano:
> I don't think we should disable the python-based one, since I see uses for that as well. I think we can explain this better to make sure there's no confusion, but it's a bit unfortunate that the names are identical.

This is good with me. We don't need anything fancy, the api is pretty different, but I think we need to be a bit more explicit in the docs.

> > (P.S. are we actually supporting SNOPT "proper" warm start?)
> 
> No we are not. Are you suggesting that we remove the "warm" option? I would be okay with that.

Yeah, we should probably remove it for now, and maybe implement it again at a later stage? From the manual it looks like we need to pass a similar set of states to the optimizer

tests/testing_utils.py
marcomangano: I think it was just inconsistent with the docstring and the expected input type, but fundamentally did not change anything - see [these lines](https://github.com/mdolab/pyoptsparse/blob/0914abf94f47e26bfbdd36cc83436d7fda5f2c04/tests/testing_utils.py#L317-L318)

sseraj: Makes sense, thanks

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: I don't think human-readable is useful, these are just a bunch of fairly long numpy arrays that are not human-readable anyways.

marcomangano: alright, let's ignore that then!

sseraj: We agreed on not changing this for now, right?

sseraj (approved):

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: Yeah, I think we should keep this as is for now, and if we want to change the API we can do that later.

PR #290, opened 2022-04-02
Added docs package requirements to setup.py

marcomangano:
## Purpose

As per title, now if the user wants to build the documentation locally, they can set all the dependencies using `pip install .[docs]`

## Expected time until merged

as soon as the tests pass?

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,9 @@ if __name__ == "__main__":
         r"""__version__ = ["']+([0-9\.]*)["']+""",
         open("pyoptsparse/__init__.py").read(),
     )[0]
+    
+    with open("doc/requirements.txt") as f:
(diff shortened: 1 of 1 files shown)


nwu63 (approved):

sseraj (commented):
Should we also prune requirements.txt (here and in other repos)? numpydoc and sphinxcontrib-bibtex are both already required by sphinx_mdolab_theme.

nwu63:
Yes that'd be a good idea. I started adding dependencies to the mdolab theme but haven't gotten around to pruning the repos.

nwu63:
I think we should also bump the patch version here.

PR #300, opened 2022-05-13
Meson Build + Proper Windows Support

jackm97:
## Purpose
Managed to get a working meson build system going. With scipy making the switch recently, I managed to draw on their work quite a bit. 

With meson out of the way I was able to get a Windows build working a lot easier and it only require altering one line of code in ``pyipoptcore.c``. So, although this PR technically gets Windows working, it's more of a byproduct of the better build system hence combining them into one PR.

## Expected time until merged
Not super urgent, but I am hoping to get a Windows build up on conda-forge sooner rather than later. I've managed to mess around with the conda-forge feedstock yml to get both Windows and Linux to build correctly. When this PR is in with a new version tag, I'll make a PR for the feedstock shortly after.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [x] Other (please describe)

New build system. Changes a lot while also changing nothing at all

## Testing
Linux passes all the tests. Windows however does not. IPOPT for ``test_rosenbrock`` fails with "92 != 0" and ``test_nsga2_multi_objective`` results in a heap corruption. For the time being I've added a platform check on both of these and warnings are issued for Windows so the tests still pass on Windows. I've ran the Windows build with my organization's software where we use IPOPT and PSQP quite heavily and all tests pass there. I'm hoping to get a conda-forge build out with a disclaimer for Windows until this is resolved.

The Windows tests also still need to be run in serial due to race conditions. We actually dealt with a similar problem in our software so I may be able to address this at another time.

## Checklist


pyoptsparse/pyIPOPT/src/callback.c
```diff
@@ -34,7 +34,9 @@
 */
 
 #include "hook.h"
-#include <unistd.h>
+#ifndef _WIN64
+#include <unistd.h>  // is this needed? not platform independent, compiles successfully w/out
```
marcomangano, line 38: This might be a leftover from older versions of IPOPT, we should test this and possibly tag a supported release.

pyoptsparse/pyNSGA2/pyNSGA2.py
```diff
@@ -4,9 +4,9 @@
 """
 # Compiled module
 try:
-    from . import nsga2  # isort: skip
+    from . import _nsga2  # isort: skip
```
marcomangano, line 7: Is Meson changing the name of the binary? Right now the tests are failing because of the import error raised on line 36

marcomangano (changes requested):
Thanks for the effort Jack! I haven't tried the Meson build myself yet, but I have a few general comments below. The main issue is that changing the name of the `nsga2` imported binary is breaking all the tests on the current Linux images.

How do you plan to maintain this build process? None of our codes run on Win so we will not generate Windows images in the foreseeable future. The only testing for this Meson build would happen when we release a new tag/version and the conda-forge repo is updated. Although I expect most of the short- and mid-term development to involve the python layer, any bugfix for Win would require some extra work on your side.

nwu63 (commented):
Before I dive into the details of the PR, I wanted to discuss some high-level ideas. A few immediate questions:
- can this work with setuptools? I think I would prefer to let setuptools keep track of the package installation, but use meson to build the extension modules. This would be a two-step build process, where meson builds the .so files and copies them to the source directory, then pip installs the package (either in-source or in whichever environment the user wants, but this should be handled transparently). This is what we do with other packages from the lab (albeit using raw Makefiles). And then we can strip out all the numpy.distutils stuff from the various `setup.py` files.
- I'm a bit confused by the use of conda here, is it doing anything beyond providing a virtual environment in Windows? You're not using the cross-platform compilation capabilities of conda right?
- The compilation of optional optimizers (where we do not distribute the source code) should be detected automatically by meson, and extension modules built if available.

jackm97:
@marcomangano 
> How do you plan to maintain this build process? None of our codes run on Win so we will not generate Windows images in the foreseeable future. The only testing for this Meson build would happen when we release a new tag/version and the conda-forge repo is updated. Although I expect most of the short- and mid-term development to involve the python layer, any bugfix for Win would require some extra work on your side.

Currently I'm working half to full time depending on the week, and I'm also a full-time grad student, so time is limited to say the least. That said pyoptsparse is essential for our code base at my job so I'm figuring out with them if I can maintain a Windows build in an official capacity. I expect this to not be a problem as we're a small team with not a lot of overhead. Worst-case, I do think I could maintain this in my free time but I'm wondering if we might be able to come up with a way of still moving releases forward without Windows in the case that I'm swamped.

> The main issue is that changing the name of the nsga2 imported binary is breaking all the tests on the current Linux images.
Without numpy disutils I had to figure out the Swig command on my own and noticed the docs said this "When linking the module, the name of the output file has to match the name of the module prefixed by an underscore" after it failing to import without the underscore. However, this morning I found the relevant command-line option to bypass that. Will commit soon

pyoptsparse/pyIPOPT/src/callback.c
jackm97: Oh this is the file I meant... not pyipoptcore. It compiles and passes on Linux without it but I don't know the code well enough to remove it.

pyoptsparse/pyCONMIN/source/cnmn00.bak
```diff
@@ -1,903 +0,0 @@
-      SUBROUTINE CNMN00 (X,VLB,VUB,G,SCAL,DF,A,S,G1,G2,B,C,ISC,IC,MS1,N1
```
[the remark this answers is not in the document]
jackm97: Oh man I'm sorry. My brain saw .bak file, which are generally auto generated backup files by text editors I thought, and deleted them out of habit of cleaning up files. Are they used? I thought they may have just got added in at one point by accident

pyoptsparse/pyNSGA2/pyNSGA2.py
jackm97: I have a fix coming in. It was due to my lack of knowledge about swig command line options

jackm97:
@nwu63 
> Before I dive into the details of the PR, I wanted to discuss some high-level ideas. A few immediate questions:
> 
>     * can this work with setuptools? I think I would prefer to let setuptools keep track of the package installation, but use meson to build the extension modules. This would be a two-step build process, where meson builds the .so files and copies them to the source directory, then pip installs the package (either in-source or in whichever environment the user wants, but this should be handled transparently). This is what we do with other packages from the lab (albeit using raw Makefiles). And then we can strip out all the numpy.distutils stuff from the various `setup.py` files.

I mostly followed the scipy repo and [these build instructions](https://scipy.github.io/devdocs/dev/contributor/meson.html). I don't see why a setuptools should be a problem but, again, I'm just not too familiar with setuptools. This is my first time working with Python build systems. I don't think scipy uses Meson yet when packaging but I could be mistaken. Setuptools is necessary for pip no? If so, it seems pretty important. If someone could point me in the right direction I could take a stab at it. I left in the legacy numpy distutils for now to maintain compatibility.

>     * I'm a bit confused by the use of conda here, is it doing anything beyond providing a virtual environment in Windows? You're not using the cross-platform compilation capabilities of conda right?

That's right. The build script itself doesn't assume a conda environment, just that all the dependency/executable paths are setup correctly so that meson can find them.

>     * The compilation of optional optimizers (where we do not distribute the source code) should be detected automatically by meson, and extension modules built if available.

I was thinking about how to best do this. I don't have these packages so I don't know their layout. Suggestions?

jackm97:
I noticed `black` failed again after trying to fix the style check errors. It seems to be a line length problem. What's your line length?

nwu63:
> I mostly followed the scipy repo and [these build instructions](https://scipy.github.io/devdocs/dev/contributor/meson.html). I don't see why a setuptools should be a problem but I'm just not too familiar with setuptools. This is my first time working with Python build systems. I don't think scipy uses Meson yet when packaging but I could be mistaken. Setuptools is necessary for pip no? If so, it seems pretty important. If someone could point me in the right direction I could take a stab at it. I left in the legacy numpy distutils for now to maintain compatibility.

Right, yeah I think we can work together to figure this out. What I'm picturing is for `meson` to build the extension modules, and to move them into the source directories. Then `setup.py` will just install everything into whatever environment is set up.

We can even have an optional testing step or something via meson, but that's secondary I think.

> That's right. The build script itself doesn't assume a conda environment, just that all the dependency/executable paths are setup correctly so that meson can find them.

Right I see, I think I would prefer documenting the build steps rather than committing the `.bat`, shell scripts, etc. into the repo.
The environment file should also be kept separate I think. Related to this, we do maintain a conda-forge package, but it doesn't support Windows yet.

> I was thinking about how to best do this. I don't have these packages so I don't know their layout. Suggestions?

There's logic in the individual `setup.py` files to check this, so I can help you port that over into the meson files.

> I noticed `black` failed again after trying to fix the style check errors. It seems to be a line length problem. What's your line length?

We use `-l 120`, see [here](https://mdolab-pyoptsparse.readthedocs-hosted.com/en/latest/contribute.html) for more info.

In terms of maintaining this, I think we can comfortably take over for Linux builds, and honestly this seems a lot more maintainable/understandable than the `numpy.distutils` stuff we had before. I don't think we will be testing Windows via our CI processes, so unless we can get the `conda-forge` package working with Windows, I'm afraid that will stay untested. Of course, if you use this regularly, then feel free to submit patches or whatever if anything breaks.

pyoptsparse/pyCONMIN/source/cnmn00.bak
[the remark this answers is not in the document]
nwu63: Yeah I don't think those are used so it's totally fine to skip. The CONMIN tests ran fine.

jackm97:
So I have a coworker that runs osx and we're going back and forth trying to get macos to build. Is there any commit message I can use to avoid wasting your CI?

jackm97:
@nwu63 I will reply more in depth to your message sometime next week. You can ignore my last message on running, I can just work on a different branch and rebase when it's all working haha. I'm a bit tired as of late, forgive me.

I've allowed edits by maintainers so feel free to add the setup.py if you want to get in before I can properly address your last comment. I agree that the meson build system is much more straight forward

nwu63:
> So I have a coworker that runs osx and we're going back and forth trying to get macos to build. Is there any commit message I can use to avoid wasting your CI?

Nah our CI is free and unlimited, don't worry about that.

> @nwu63 I will reply more in depth to your message sometime next week. You can ignore my last message on running CI, I can just work on a different branch and rebase when it's all working haha. I'm a bit tired as of late, forgive me.
> 
> I've allowed edits by maintainers so feel free to add the setup.py if you want to get in before I can properly address your last comment. I agree that the meson build system is much more straight forward

Sure no worries, I'll push up some stuff this week and we can iterate.

jackm97:
> Sure no worries, I'll push up some stuff this week and we can iterate.
Great! Looking forward to a more painless build process across platforms. Btw I forked your conda feedstock for my conda-build testing and was making good progress. We may be building in-house soon to take advantage of snopt so we would put our Windows CI to work there and help catch bugs in tandem with the the conda-forge CI.

Another option I thought of is to maintain a separate conda-forge feedstock for windows, like pyoptsparse-win64 or something, for now. I imagine there will be a fair amount of trial and error there and I wouldn't want it to conflict with the progress of your feedstock

pyoptsparse/pyCONMIN/source/cnmn00.bak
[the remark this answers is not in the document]
eli-schwartz: I would typically do something like that in its own commit with a commit message such as "delete unused files", which would make it a lot clearer what is going on. :)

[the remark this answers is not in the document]
jackm97: It did have it's own commit and then I rebased... poor choice of a squash on my end haha. Noted for future reference

pyoptsparse/pyNSGA2/pyNSGA2.py
jackm97: Fixed

eli-schwartz:
> * can this work with setuptools? I think I would prefer to let setuptools keep track of the package installation, but use meson to build the extension modules. This would be a two-step build process, where meson builds the .so files and copies them to the source directory, then pip installs the package (either in-source or in whichever environment the user wants, but this should be handled transparently). This is what we do with other packages from the lab (albeit using raw Makefiles). And then we can strip out all the numpy.distutils stuff from the various `setup.py` files.



> I mostly followed the scipy repo and [these build instructions](https://scipy.github.io/devdocs/dev/contributor/meson.html). I don't see why a setuptools should be a problem but I'm just not too familiar with setuptools. This is my first time working with Python build systems. I don't think scipy uses Meson yet when packaging but I could be mistaken. Setuptools is necessary for pip no? If so, it seems pretty important. If someone could point me in the right direction I could take a stab at it. I left in the legacy numpy distutils for now to maintain compatibility.

setuptools is not necessary for pip, pip expects one of the following:
- a PEP 517 build backend: https://peps.python.org/pep-0517/
- the implicit / default setuptools build backend

SciPy intends to use an alternative build backend which directly uses Meson and fully drop any and all references to setuptools, although currently during the transition period the default pip install experience uses setuptools and doesn't invoke Meson at all.

nwu63:
@eli-schwartz I have a few questions regarding meson, apologies in advance since I'm not super familiar with it
- I'm still a bit confused about the interaction between meson and python packaging. The standard approach for installing in meson will just install it into the environment that's detected by meson right? When I tried this, `pip` was not aware of this installed package. Is there a way for pip to manage the installation (e.g uninstall using pip), get the version, etc.?
- More broadly speaking, I'm still more used to the setup.py approach where an in-source installation is allowed for development purposes (for example via `pip install -e`). Is there a way to replicate this behaviour? I guess an alternative is to somehow get `meson` to copy the `.so` files (which are really the only ones that need to be built by meson) to the source dir (where they currently live using numpy.distutils), but my suspicion is that this is not allowed in meson by design. How do most python packages get around this?

jackm97:
> @eli-schwartz I have a few questions regarding meson, apologies in advance since I'm not super familiar with it
> 
>     * I'm still a bit confused about the interaction between meson and python packaging. The standard approach for installing in meson will just install it into the environment that's detected by meson right? When I tried this, `pip` was not aware of this installed package. Is there a way for pip to manage the installation (e.g uninstall using pip), get the version, etc.?
> 
>     * More broadly speaking, I'm still more used to the setup.py approach where an in-source installation is allowed for development purposes (for example via `pip install -e`). Is there a way to replicate this behaviour? I guess an alternative is to somehow get `meson` to copy the `.so` files (which are really the only ones that need to be built by meson) to the source dir (where they currently live using numpy.distutils), but my suspicion is that this is not allowed in meson by design. How do most python packages get around this?

@nwu63 I know this wasn't directed at me but I think one option for maintaining backwards compatibility with `setuptools` is to first build and install the package into the current directory with meson at `./installdir` or something. This would handle all the copying of .so files and such. We can add the `setup.py` files to the meson sources so they get copied over as well. Then we can run `setup.py` in the install directory for `pip`. This would unfortunately require a build step outside of the `setup.py` call which may not be okay. I'm not familiar with `setuptools` haha. I prefer `conda-build` as it supports local build and installs and has access to conda-forge which is just packed with software.

We could probably also run the meson build and install to `$PWD/installdir` from within `setup.py` but idk how that'd work with the moving of sources.

jackm97:
How does conda-forge work for contributing to recipes? I have a fork of your feedstock and managed to get it to build on Windows and Linux. I'm wondering what the process for submitting a PR to the feedstock is after this is all merged

nwu63:
> @nwu63 I know this wasn't directed at me but I think one option for maintaining backwards compatibility with `setuptools` is to first build and install the package into the current directory with meson at `./installdir` or something. This would handle all the copying of .so files and such. We can add the `setup.py` files to the meson sources so they get copied over as well. Then we can run `setup.py` in the install directory for `pip`. This would unfortunately require a build step outside of the `setup.py` call which may not be okay. I'm not familiar with `setuptools` haha. I prefer `conda-build` as it supports local build and installs and has access to conda-forge which is just packed with software.
> 
> We could probably also run the meson build and install to `$PWD/installdir` from within `setup.py` but idk how that'd work with the moving of sources.

I'm fine with a two-step build process (first compile the extension modules, then run pip) since this is what we do with most of our packages other than pyOptSparse anyhow. But neither of these approaches would really enable an editable installation in the pythonic sense (while keeping the source files under git control, which is nice for a variety of reasons). When I am developing in pyOptSparse, I want the compiled extension modules in my source tree so that I can debug and commit as necessary, without having to check whether I am running the python files that I think I'm running.

One possible solution is to somehow re-use the `build` directory from pip, by placing the .so files in the correct location so that `pip` will pick them up and copy them to the installation directory (either in-source, as indicated by `-e`, or in site-packages). I don't really know the inner workings of either pip/meson, but perhaps this is possible.

Unfortunately, while conda has many nice features especially for Windows, it causes too many problems for our use cases (and at the end of the day we prioritize the use cases here at the MDO Lab, which is for development under Linux). Plus, we can't really use it on HPC systems.

> How does conda-forge work for contributing to recipes? I have a fork of your feedstock and managed to get it to build on Windows and Linux. I'm wondering what the process for submitting a PR to the feedstock is after this is all merged

I think you can just open a PR, and iterate with the CI there until things pass.

jackm97:
@nwu63 
> I'm fine with a two-step build process (first compile the extension modules, then run pip) since this is what we do with most of our packages other than pyOptSparse anyhow. But neither of these approaches would really enable an editable installation in the pythonic sense (while keeping the source files under git control, which is nice for a variety of reasons). When I am developing in pyOptSparse, I want the compiled extension modules in my source tree so that I can debug and commit as necessary, without having to check whether I am running the python files that I think I'm running. And we definitely don't want to achieve this via `PYTHONPATH`.
> 
> 
> 
> One possible solution is to somehow re-use the `build` directory from pip, by placing the .so files in the correct location so that `pip` will pick them up and copy them to the installation directory (either in-source, as indicated by `-e`, or in site-packages). I don't really know the inner workings of either pip/meson, but perhaps this is possible.

Okay knowing your specific workflow helps a lot. I have an idea and will try it out when I get home. 

> 
> Unfortunately, while conda has many nice features especially for Windows, it causes too many problems for our use cases (and at the end of the day we prioritize the use cases here at the MDO Lab, which is for development under Linux). Plus, we can't really use it on HPC systems.

It's funny, with all my emphasis on Windows, I loathe developing for Windows haha. I'm a Linux user through and through. Everyone that works on our software runs MacOS or Linux but everyone that uses our software runs Windows so here we are. I know conda got a lot of flak years back because of issues mixing conda-forge and anaconda official repos but a few versions back they finally implemented a strict channel priority. I've been running conda for quite a while on Linux without a hitch and I absolutely love it. Not trying to convince you out of your workflow tho! I just really like conda haha. Also I know next to nothing about HPC workflows. 

> I think you can just open a PR, and iterate with the CI there until things pass. 

Good to know. So the CI runs on PRs?

nwu63:
> Okay knowing your specific workflow helps a lot. I have an idea and will try it out when I get home.

Thanks for your help btw! I know there's a big difference between end users and developers, in particular for build systems.

> It's funny, with all my emphasis on Windows, I loathe developing for Windows haha. I'm a Linux user through and through. Everyone that works on our software runs MacOS or Linux but everyone that uses our software runs Windows so here we are. I know conda got a lot of flak years back because of issues mixing conda-forge and anaconda official repos but a few versions back they finally implemented a strict channel priority. I've been running conda for quite a while on Linux without a hitch and I absolutely love it. Not trying to convince you out of your workflow tho! I just really like conda haha. Also I know next to nothing about HPC workflows.

Yeah I have nothing positive to say about developing on Windows :stuck_out_tongue:  My main gripe with conda is that it really wants you to use its entire toolchain for everything, from compilers to linkers etc. If you're fully immersed in the conda ecosystem and you're using it to build the whole thing, then it will work pretty nicely (and on multiple architectures), but if you don't want to do it a certain way, then you run into all sorts of problems. Since we are mainly developers and don't really do binary distributions, there's very little benefit for us compared to something like pyenv. Regarding HPC, there are certain packages that conda will want to vendor and have you use, but we need to use what's provided by the HPC and it's tricky to get conda to behave in that situation.

> Good to know. So the CI runs on PRs?

Yep, I believe so. Windows builds are currently disabled but we can change that.

jackm97:
> Thanks for your help btw! I know there's a big difference between end users and developers, in particular for build systems.
Of course! This is one of the first times I actually have the ability to contribute to a package I use. It feels good haha

> Yeah I have nothing positive to say about developing on Windows stuck_out_tongue My main gripe with conda is that it really wants you to use its entire toolchain for everything, from compilers to linkers etc. If you're fully immersed in the conda ecosystem and you're using it to build the whole thing, then it will work pretty nicely (and on multiple architectures), but if you don't want to do it a certain way, then you run into all sorts of problems. Since we are mainly developers and don't really do binary distributions, there's very little benefit for us compared to something like pyenv. Regarding HPC, there are certain packages that conda will want to vendor and have you use, but we need to use what's provided by the HPC and it's tricky to get conda to behave in that situation.
Hmm yeah that makes total sense. It's definitely geared towards an isolated environment from the system binaries/libraries. 

> Yep, I believe so. Windows builds are currently disabled but we can change that.
Awesome. I may start playing with it

eli-schwartz:
@nwu63

Meson doesn't currently include a mechanism to generate pip metadata (the .dist-info directory) so you need a pep517 build backend to run pip and do that. This is on my long-term list of improvements to make, but... there's a snag. Meson has a strict no-dependencies policy because it's used to bootstrap operating systems by building very low-level C components, and that becomes harder the more things you need to add to the bootstrap chain... and python packaging isn't a good mix there. (I have experience trying to bootstrap python packaging for a distro, and it's a big problem even for a fully built distro just to bootstrap a new version of python. No one wants that struggle while they are in the middle of bringing up the base system.) Meson actually supports several installation methods, two of which do not use pip or anything else that would cause issues; it's also possible if users are okay with requiring setuptools before meson, to use that.

Unfortunately, to do any kind of python packaging metadata, we need a toml parser in the stdlib. The python developers dragged their feet on agreeing to the need for this for multiple development cycles of python, but one finally did get added in python 3.11 which is not yet released. Meson still supports python 3.7. It's not clear when or how we can start using it, it's barely possible we could tell people to use python 3.11 if they don't want to deal with dependencies but then we still have to wait for 3.11 to be released and then for it to be broadly available...

Until then, meson can do all the build+install as well as uninstall on its own, but not create the metadata pip needs to do an uninstall. That's why pep517 build backends exist. You'd then `pip install .` and pip would run meson behind the scenes, process the built output into a python wheel, and install the wheel (and uninstall it when you want, too).

As time goes by, Meson *will* grow to do all this for you.

...

Regarding setuptools `develop` / `pip install -e` the build backend will need to support this. It's possible to do it directly in Meson by using `meson.add_devenv()` to set up PYTHONPATH but of course that has its own challenges (particularly combining pure python and compiled extensions into one layout).

jackm97:
My most recent commit has `setuptools` support! The old build instructions should work just fine. I still have some Meson build instruction notes in the README. I'll clean it all up a bit more once I get windows working

nwu63:
@eli-schwartz Thanks for the info, I did follow the situation a little bit on GitHub, and I totally understand the difficulty in implementing something like this in meson. At this moment I'm in favour of letting pip handle the installation, but I don't know if it's a good idea to 1) call `meson` via `os.system` in Python, and 2) use `shutil` to copy the extensions into the Python source tree. Is there a way to specify in meson the installation directory to be the Python source directory, such that the install step from meson would just copy the .so files into the Python source tree. Then we can call `pip` to install as before.

@jackm97 Thanks for the implementation, it seems to work well, but I have some questions/slight reservations (see above). I know that for some of the other packages we will definitely want to call `meson` manually so that it's more configurable.

eli-schwartz:
You can invoke Meson with `-Dpython.purelibdir=path/to/install/all/dot-py/files -Dpython.platlibdir=path/to/install/compiled/extensions`

jackm97:
> You can invoke Meson with `-Dpython.purelibdir=path/to/install/all/dot-py/files -Dpython.platlibdir=path/to/install/compiled/extensions`

I tried this and while I was able to simplify some things by not having a to deal with a bunch of nested directories in the install prefix, trying to set the install prefix to the repo directory did not seem to work. I don't really know why but I do know the meson install command kept deleting `pyoptsparse/__init__.py`

nwu63 (commented):
Hm I think the approach via `os.system` (or maybe using `subprocess` would be slightly better) is a lot more explicit/readable than the current setup. Also see below for a few comments.

nwu63:
One more thing: I'm having trouble getting things compiled correctly, getting
```
ImportError: numpy.core.multiarray Failed to Import
```
when I try to import any of the compiled libraries. Looking through the log, it seems that due to the presence of `pyproject.toml`, everything is built from a temporary directory so the f2py/numpy includes point to locations that no longer exist. I'm not sure what exactly is going on, but if I remove the `pyproject.toml` file then everything works fine for me. Are you seeing the same thing on your end @jackm97?

.github/build_real.sh
```diff
@@ -7,4 +7,5 @@ if [[ $IMAGE == "private" ]]; then
     cp -r $HOME/SNOPT/* pyoptsparse/pySNOPT/source
 fi
 
-pip install .[optview,testing]
+pip install --upgrade numpy
```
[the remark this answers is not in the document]
jackm97: This actually related to your comment:
> when I try to import any of the compiled libraries. Looking through the log, it seems that due to the presence of pyproject.toml, everything is built from a temporary directory so the f2py/numpy includes point to locations that no longer exist. I'm not sure what exactly is going on, but if I remove the pyproject.toml file then everything works fine for me. Are you seeing the same thing on your end @jackm97?

It seems `pip install .` doesn't check the numpy version of the environment it's installing to. It just installs the latest `numpy` version for the build. So once the package is installed and used it fails because it was built with a different version of `numpy` than what is installed in the target environment

README.md
```diff
@@ -39,6 +39,24 @@ Please see the [documentation](https://mdolab-pyoptsparse.readthedocs-hosted.com
 Testing is done with the `testflo` package developed by the openMDAO team, which can be installed via `pip install testflo`.
 To run the tests, simply type `testflo .` in the root directory.
 
+## Meson Build Instructions
```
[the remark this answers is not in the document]
jackm97: Yeah I know :/ I mentioned somewhere in the thread I'd clean it up. It's mostly notes for myself currently. Sorry about that

pyoptsparse/pyIPOPT/find_python_path.py
```diff
@@ -0,0 +1,14 @@
+# Standard Python modules
```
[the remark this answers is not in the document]
jackm97: Oh yeah it's from an earlier iteration. It can be removed. I'll resolve it in a sec

setup.py
```diff
@@ -90,7 +130,12 @@ def configuration(parent_package="", top_path=None):
             "Topic :: Software Development",
             "Topic :: Education",
         ],
-        configuration=configuration,
+        package_dir={"": "."},
+        packages=setuptools.find_packages(where="."),
+        package_data={
+            "": ["*.so", "*.lib", "assets/*"],
+        },
+        python_requires=">=3.8",
```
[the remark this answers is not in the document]
jackm97: I wasn't sure. I just used the same python version as your azure pipeline

.github/build_real.sh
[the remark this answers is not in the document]
eli-schwartz: https://numpy.org/devdocs/user/depending_on_numpy.html#build-time-dependency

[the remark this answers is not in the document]
jackm97: In that case it seems we should use `==1.16` instead of `>=1.16` for numpy?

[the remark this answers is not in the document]
jackm97: It seems the meson build doesn't like that...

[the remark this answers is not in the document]
jackm97: or the tests on my local machine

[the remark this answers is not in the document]
jackm97: I think I managed to fix it with [10ba333](https://github.com/mdolab/pyoptsparse/pull/300/commits/10ba3330066767b53b70d2364d0659e5a01c181c)

pyoptsparse/pyIPOPT/find_python_path.py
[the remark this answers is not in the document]
jackm97: resolved with [73b5ff0](https://github.com/mdolab/pyoptsparse/pull/300/commits/73b5ff0bdd953f4b9d2b2adec14080e01f8a54a8)

README.md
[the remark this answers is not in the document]
jackm97: Moved build instructions to rst

pyoptsparse/meson.build
```diff
@@ -0,0 +1,88 @@
...
+python_paths = [join_paths(meson.current_build_dir(), '..')]
+envdata.prepend('PYTHONPATH', python_paths)
+
+progs = [['SLSQP',  'pySLSQP',  'slsqp'],
+         ['CONMIN', 'pyCONMIN', 'conmin'],
+         ['PSQP',   'pyPSQP',   'psqp'],
+         ['NSGA2',  'pyNSGA2',  'nsga2']]
+
+
+foreach p : progs
```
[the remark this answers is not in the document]
jackm97: I'm not sure this will test if they got built correctly. The `meson.build` files only get parsed on the setup step if I'm not mistaken. Unless I'm misunderstanding what you're saying

setup.py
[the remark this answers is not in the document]
jackm97: 3.7 works fine. changed it back

jackm97:
Okay I'm feeling pretty good about the latest series of commits. They are mostly related to conda-forge stuff (you may have seen I submitted a PR to the feedstock as well)

I can't really think of anything else to work in there so I'll be done committing now. Sorry for all the commit spam! A lot of trial and error with this one, especially after running the CI haha.

Let me know if you have any more concerns! I may add a commit or two after the results of the latest CI jobs come in

pyoptsparse/pyIPOPT/meson.build
```diff
@@ -0,0 +1,58 @@
+fs = import('fs')
+
+if get_option('ipopt_dir') != '' or fs.is_dir('Ipopt')
```
nwu63, line 3: In addition to the `IPOPT_DIR` approach, the original `setup.py` had several other ways of detecting the library:
- searching for a directory under `pyIPOPT`
- allowing for specifying include and lib dirs separately via `IPOPT_INC` and `IPOPT_LIB`

jackm97:
> Thanks for the work, some more questions:
> 
>    * if we are only using `meson` to build the shared libraries, shouldn't we remove all the `install_sources` for the Python stuff? We only really care about the final libraries right? Similarly, we probably don't need a `meson.build` file for directories that aren't compiling anything, e.g. `ALPSO` or `postprocessing`.

There are cases where workflows not dealing with setup.py are easier imo, especially with getting the conda-forge package working with cross-compiling osx-arm64. The same `distutils` trickery conda-forge employs doesn't work with meson from what I was trying. Even without conda-forge, there's something to be said about just running:
```shell
meson setup build --prefix=/path/to/install
meson install -C build
```
Setuptools just simply adds metadata for the pip installation at this point, so if you don't care about that, it's nice to have the flexibility of only dealing with `meson`. In this case we'd want all the source files, including `.py` and the stuff in `postprocessing` to get sent to the desired installation path. Plus, `meson` will probably be PEP517 compliant in the near future, allowing us to use it as the build-backend and maybe even a front-end as well. If/when this happens it'll be nice to already have the meson.build files contain all sources. I think your following remark is a good compromise there.

> 
>    * Do we still need `staging_dir`? Can we just copy the libraries from the build dir directly to the source tree?

Probably not. I don't see why we have to run the install step from `setup.py`. We could just compile the .so files into `meson_build` and copy. I actually tried doing this but ran into a problem and got annoyed so just left it. I think the problem was me being tired haha.

> 
>    * There are a few minor things that I was thinking of changing, but they are unimportant and we can figure this stuff out later:
>       
>        * rename the libraries to follow `lib<package>.so` naming convention
>        * I don't know if it makes sense to have a dedicated `lib` directory where we can just move things to, instead of having them live in the subdirs. If we do this, I think we no longer need to use `shutil` in `setup.py` to copy stuff, but can directly set the installation directory to the libdir following `-Dpython.platlibdir` when calling meson.

I think maybe these should be options if/when they get implemented as they go against standards no?

pyoptsparse/meson.build
```diff
@@ -0,0 +1,106 @@
+# Platform detection
+is_windows = host_machine.system() == 'windows'
+is_msvc = is_windows and cc.get_id() == 'msvc'
+
+# NumPy include directory - needed in all submodules
+incdir_numpy = get_option('incdir_numpy')
```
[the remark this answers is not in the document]
jackm97: There were some extra compiler flags for Windows initially. It can be removed.

pyoptsparse/pyIPOPT/meson.build
jackm97: Ugh... another annoying meson particularity. It finds the IPOPT library automatically, but then you can't get that path so you cant find the header...

jackm97: I'm actually going to try something real quick. I have an idea for how to handle this

jackm97: I'm realizing I misread your comment... searching `pyIPOPT` directory was already supported. I did push up a commit that cleaned up the logic a bit though. `IPOPT_INC` and `IPOPT_LIB` are not however, I can add that back in soon, but is there a use case where the headers and shared library don't follow the layout `$IPOPT_DIR/coin{,-or}` and `$IPOPT_DIR/lib`?

meson.build
```diff
@@ -0,0 +1,43 @@
...
+# We need -lm for all C code (assuming it uses math functions, which is safe to
+# assume for SciPy). For C++ it isn't needed, because libstdc++/libc++ is
+# guaranteed to depend on it. For Fortran code, Meson already adds `-lm`.
+m_dep = cc.find_library('m', required : false)
+if m_dep.found()
+  add_project_link_arguments('-lm', language : 'c')
+endif
+
+# Adding at project level causes many spurious -lgfortran flags.
+add_languages('fortran', native: false)
```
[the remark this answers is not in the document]
nwu63: Yeah that's a fair reason :P

pyoptsparse/pyIPOPT/meson.build
nwu63: If the user installs IPOPT via yum on CentOS for example, I think the system lib dir is typically `/usr/lib64/` instead of `/usr/lib/`. We can of course add in some sort of smart detection, but allowing the user to specify both independently may be a more flexible solution.

pyoptsparse/pyIPOPT/src/callback.c
nwu63: Yeah I'm not sure, but if all tests pass without it then I think we can get rid of it.

tests/test_rosenbrock.py
```diff
@@ -150,6 +153,15 @@ def test_snopt_hotstart_starting_from_grad(self):
 
     @parameterized.expand(["IPOPT", "SLSQP", "PSQP", "CONMIN", "NLPQLP", "ParOpt"])
     def test_optimization(self, optName):
+        try:
+            self._test_optimization(optName)
+        except AssertionError as e:
+            if optName == "IPOPT" and sys.platform == "win32":
```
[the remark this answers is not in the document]
nwu63: I think we can do this via a unittest decorator, e.g.
```
@unittest.skipIf(optName == "IPOPT" and sys.platform == "win32", reason)
```

but I don't know 100% if this is compatible with what we currently do. I don't have a way to test this on Windows, but if it works that would be cleaner than the current setup.

pyoptsparse/pyIPOPT/meson.build
nwu63: Actually, I see that you already handle lib64 here. Maybe that's okay then, if we find a use case we can always add this feature back.

nwu63:
> There are cases where workflows not dealing with setup.py are easier imo, especially with getting the conda-forge package working with cross-compiling osx-arm64. The same `distutils` trickery conda-forge employs doesn't work with meson from what I was trying. Even without conda-forge, there's something to be said about just running:
> 
> ```shell
> meson setup build --prefix=/path/to/install
> meson install -C build
> ```
> 
> Setuptools just simply adds metadata for the pip installation at this point, so if you don't care about that, it's nice to have the flexibility of only dealing with `meson`. In this case we'd want all the source files, including `.py` and the stuff in `postprocessing` to get sent to the desired installation path. Plus, `meson` will probably be PEP517 compliant in the near future, allowing us to use it as the build-backend and maybe even a front-end as well. If/when this happens it'll be nice to already have the meson.build files contain all sources. I think your following remark is a good compromise there.

I think the philosophy of meson, while it makes a lot of sense for compiled projects, doesn't work too well with Python. Having to list individual Python files manually is pretty error-prone, and I don't think any of the maintainers can guarantee that this installation method is going to be maintained going forward.

> Probably not. I don't see why we have to run the install step from `setup.py`. We could just compile the .so files into `meson_build` and copy. I actually tried doing this but ran into a problem and got annoyed so just left it. I think the problem was me being tired haha.

Yeah I'd prefer if `staging_dir` is removed to simplify the code/build process for future maintainability.

> > * There are a few minor things that I was thinking of changing, but they are unimportant and we can figure this stuff out later:
> >   
> >   * rename the libraries to follow `lib<package>.so` naming convention
> >   * I don't know if it makes sense to have a dedicated `lib` directory where we can just move things to, instead of having them live in the subdirs. If we do this, I think we no longer need to use `shutil` in `setup.py` to copy stuff, but can directly set the installation directory to the libdir following `-Dpython.platlibdir` when calling meson.
> 
> I think maybe these should be options if/when they get implemented as they go against standards no?

I guess it varies in the Python world. In any case it's not a big deal.

doc/install.rst
```diff
@@ -62,8 +64,15 @@ It is also possible to install pyOptSparse by calling ``python setup.py install`
 .. note::
   Some optimizers are proprietary and their sources are not distributed with pyOptSparse.
   To use them, please follow the instructions on specific optimizer pages.
-  
-For those who intend to use pyOptSparse with IPOPT, OpenMDAO developers provide a `bash script <https://github.com/OpenMDAO/build_pyoptsparse>`_ that simplifies the installation of the optimizer with different linear solvers.
+
+For those who intend to use pyOptSparse with IPOPT, OpenMDAO developers provide a `bash script <https://github.com/OpenMDAO/build_pyoptsparse>`_ that simplifies the installation of the optimizer with different linear solvers. With IPOPT installed, set the environment variable ``IPOPT_DIR`` to the path of the IPOPT installation. The installation directory structure is expected to look like the following:
```
nwu63, line 68: The IPOPT instructions should be moved to the IPOPT page, and merged with the content there.

doc/environment.yml
```diff
@@ -0,0 +1,18 @@
+dependencies:
```
nwu63, line 1: I don't think we should maintain these `enviornment.yml` files, the information here is basically the combination of what's in `setup.py` and `pyproject.toml`, and I just don't think anyone will be updating these over time since none of us use Windows or conda.

I noticed that you needed `libpgmath` for Windows, that sort of information may be better suited in the docs.

meson.build
```diff
@@ -0,0 +1,45 @@
+# Much of this is from SciPy
+
+project(
+  'pyoptsparse',
+  'c', 'cpp',
+  # Note that the git commit hash cannot be added dynamically here (it is added
```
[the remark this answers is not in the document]
jackm97: > I think the philosophy of meson, while it makes a lot of sense for compiled projects, doesn't work too well with Python. Having to list individual Python files manually is pretty error-prone, and I don't think any of the maintainers can guarantee that this installation method is going to be maintained going forward.

I quote this comment because it's related. This is technically a compiled project and the new Python module in Meson is specifically for dealing with packages that require extension modules. I'd argue that `setuptools` doesn't make sense for this project *because* of the compiled extension modules. While not available yet, with Scipy using meson as their build system, we will see a PEP517 interface that allows one to set `meson` as the build backend in the `pyproject.toml` so `pip` works naturally. It's on their list of things to do mentioned in this [article](https://scipy.github.io/devdocs/dev/contributor/meson.html). Ideally, at some point, `setuptools` will be dropped from the build process because right now this PR has a build backend calling a build backend which is not great. Once this happens, `meson` will be the one providing metadata to pip, not `setuptools`.

In summary, things are weird right now, but there should come a day soon where meson can be a replacement for `setuptools` as a backend. This means maintaining 
```
meson setup build && meson -C build install
```
and maintaining `pip install .` are one in the same. No extra work.

That said, using `meson` as a backend isn't supported yet so we could revisit this again once in the future and instead just use meson for the compiled extension modules for now. But I just wanted to get it out there that the Python community formerly using `numpy.distutils` is trending towards `meson` as a build backend which means the workflow of `pip` remains the same, it's just a different backend.

meson_options.txt
```diff
@@ -0,0 +1,6 @@
+option('ipopt_dir', type: 'string', value: '',
+        description: 'Top-level dir for ipopt')
+option('incdir_numpy', type: 'string', value: '',
```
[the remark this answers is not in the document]
jackm97: Yeah they are. At least for cross-compilation, which is necessary for osx-arm64 builds on conda-forge. Cross-compilation took some time to get working so I'd really like if this could stay in! It doesn't really require any extra maintenance on your end but the option is really nice on my end where cross-compilation would be super helpful for our CI.

doc/install.rst
jackm97: Sounds good. Will get to it sometime next week I think

meson.build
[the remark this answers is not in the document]
jackm97: [1f56a44](https://github.com/mdolab/pyoptsparse/pull/300/commits/1f56a444c5b7f68574e685f3ee149707a4eab42c) attempts to resolve this. I've removed(commented out) everything from `meson.build` whose functionality is currently duplicated by `setuptools` to help with maintainability on your end until `meson` gets a PEP517 interface!

doc/environment.yml
jackm97: Okay. I'm going to try beefing up the requirements section in the docs then if that's alright. When looking to build the package from source my first time, it was a little confusing having requirements spread across the install instructions. An `environment.yml` is really nice for my sanity but I understand the maintenance burden

nwu63: Yeah I know, it's just that we don't really use conda and I hesitate to officially support it, especially the more complicated stuff with cross-compilation. I was hoping that between the docs and the conda-forge package that would cover most people's needs.

meson.build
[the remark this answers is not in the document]
nwu63: Yeah that's fair, and I can see us ditching pip if meson works that way. I'm not super familiar with all the python build/packaging options out there and the ecosystem is a bit of a mess. Since all the maintainers here are just PhD students, I really want to make sure that whatever we do accept is 1) understandable and maintainable by us, and 2) useful for everyone, but with priority given to those in the lab and our collaborators. That's why I'm pretty sold on switching from `numpy.distutils` to `meson`, but less so for `conda` and Windows builds, since our ecosystem doesn't really rely on those. In any case, thanks for being patient with my questions, and it is nice to know that people are not only using our FOSS but also willing to contribute to it.

meson_options.txt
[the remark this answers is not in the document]
nwu63: Okay, as long as we have some sort of documentation (maybe just as comments here, e.g. what would be typical values passed in) I'm fine with keeping those. I don't see us touching these options so I want to at least get an idea of how to use them.

meson.build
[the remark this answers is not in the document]
jackm97: As a Master's student myself, and soon to be PhD student, I can empathize there haha. Just to be clear `pip` would still be the frontend for you all. I think @eli-schwartz may have linked it somewhere already, but I recommend reading [PEP517](https://peps.python.org/pep-0517/) if you haven't already. Essentially the standard opens up the build backend to anything that implements the standard, rather than locking `pip` down to `setuptools` and the like.

I will be continuously testing Windows builds for work, and since you're code doesn't do any crazy platform specific stuff, it should be fairly easy. If I come across problems on Windows I will open pull requests if I can, but definitely no expectation for Windows maintenance on your end haha. 

With `conda`/cross-compiling, it's a similar story. Except for the separation of `python_target`/`python_command` in `meson.build`, the actual cross-compilation burden is offset to those who want to use it. So I don't think it'll be  a maintenance burden on your end since `python_target` and `python_command` are the same unless otherwise specified. I don't expect the meson api for getting the python installation will change much, if at all. Also, important note, without the separation of `python_target`/`python_command`, your conda-forge package runs into problems. But yeah, I don't intend to add cross-compilation to your list either haha. Just a request that some flexibility, if simple, remains so that exotic use cases like cross-compilation are available to the user.

Finally, another use case of `python_target`/`python_command` that may be of more interest to you. Meson has Python as a dependency and by default, `py3 = py_mod.get_installation()` grabs the Python version used to run Meson. So consider the case that you use `meson` installed on your system using `apt` for example, but you want to build and install `pyoptsparse` for the Python version in your `pyenv`. Without separating `python_target`/`python_command`, this would not be possible because  `py3 = py_mod.get_installation()` would grab your system installation, not what's in `pyenv`. By separating them you can now do `py3_command = py_mod.get_installation()` for getting the system installation associated with `meson` and  `py3_target = py_mod.get_installation('/path/to/pyenv/bin/python')` to install to your `pyenv` installation.

meson_options.txt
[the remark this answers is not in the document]
jackm97: > Finally, another use case of python_target/python_command that may be of more interest to you. Meson has Python as a dependency and by default, py3 = py_mod.get_installation() grabs the Python version used to run Meson. So consider the case that you use meson installed on your system using apt for example, but you want to build and install pyoptsparse for the Python version in your pyenv. Without separating python_target/python_command, this would not be possible because py3 = py_mod.get_installation() would grab your system installation, not what's in pyenv. By separating them you can now do py3_command = py_mod.get_installation() for getting the system installation associated with meson and py3_target = py_mod.get_installation('/path/to/pyenv/bin/python') to install to your pyenv installation.

This part of my previous comment answers your question about `python_target` I think. But I will add some notes in comments later and resolve this once I do

doc/environment.yml
jackm97: That makes total sense. Yeah I'll remove it and add some docs for the requirements!

tests/test_rosenbrock.py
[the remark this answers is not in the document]
jackm97: I will try it. Where would I put it?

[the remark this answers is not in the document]
nwu63: I realized what I said above probably won't work. I pushed up something slightly different, maybe you can try it out and see if it works.

pyoptsparse/pyIPOPT/meson.build
jackm97: Is it okay to resolve this?

nwu63: Yep!

doc/environment.yml
jackm97: Do we want to leave in one environment.yml for Windows CI?

pyoptsparse/meson.build
```diff
@@ -0,0 +1,106 @@
...
+      ],
+      check: true
+    ).stdout().strip()
+endif
+incdir_numpy = '''@0@'''.format(incdir_numpy)
+
+# Meson prefixes filenames of intermediate compiled objects with their filepath. This poses a problem for conda builds
+# since conda ensures the host environment directory has 255 characters so the meson object filenames then exceed 255
+# characters. To remedy this, the fortranobject.c file from numpy is copied into pyoptsparse so that the meson build
+# uses a relative path, rather than an absolute path, thus reducing the auto generated object filename
```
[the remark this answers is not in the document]
nwu63: Yep

doc/environment.yml
nwu63: Yeah if it's needed for the CI then we can leave it in. However, I would prefer if it goes under the `.github` directory instead of its current location.

tests/test_rosenbrock.py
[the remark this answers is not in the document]
jackm97: It works!

doc/environment.yml
jackm97: Will do. I'll move it when I add the CI stuff for Windows

jackm97: @nwu63 So I'm not sure how to get Windows CI working. Below is the Azure job I was going to add, but idk where to add it

```yaml
- job: Windows
    pool:
      vmImage: windows-latest

    - powershell: Write-Host "##vso[task.prependpath]$env:CONDA\Scripts"
      displayName: Add conda to PATH

    - script: conda config --add channels conda-forge && conda config --set channel_priority strict
      displayName: Set channel priority

    - script: conda create --yes --name pyos-build
      displayName: Create environment

    - script: |
        call activate pyos-build
        call conda install -y mamba
        call mamba env update --file .github/environment.yml
        call mamba install -y libpgmath
      displayName: Install mamba and update environment

    - script: |
        set IPOPT_DIR=%CONDA_PREFIX%\Library
        set CC=cl
        set FC=flang
        set CC_LD=link
        python -m build -n -x .
        pip install --no-deps --no-index --find-links dist pyoptsparse
      displayName: Build and install pyoptsparse

    - script: |
        cd tests
        testflo -n 1 .
      displayName: Run tests
```

nwu63: Hmm so this is a bit tricky, because as you see we use templates for our Pipelines. I'm not sure what's the easiest way to add a separate, standalone job to the existing stuff, do we need a separate YAML file and a separate Pipeline? Perhaps it's easiest if we use GitHub actions instead?

jackm97: I tried adding a GitHub Action and it says it needs to be approved. Hoping that works! Also, I accidentally merged main into meson instead of rebasing the meson branch onto main. Let me know if you want to me to clean the history up

sseraj: I just approved the workflow. Also, we squash the commits, so merging is fine.

jackm97: Sounds good. Looks like any change I make requires a new approval...

nwu63: I've temporarily adjusted permissions so that it should not require new approval. Also, maybe we can use [this action](https://github.com/marketplace/actions/setup-miniconda) to set up miniconda in GHA?

jackm97: Ah yes thank you. I found that a few days ago but couldn't find it again today. I will make that change

jackm97: Windows tests pass! What from here?

nwu63: Nice! Two minor things for the GHA:
1. I think it'd be nice to have `--pre_announce -v` added to the testflo command for Windows so that we can see which tests were skipped
2. Is it possible to install IPOPT via conda on Windows? That way we can test the IPOPT wrapper also

jackm97: Great! Once this PR is into main, I will update the conda-forge feedstock to point to main and to have the correct version. Regarding your comments:

1. I can add this when I'm back at my desk. 
2. IPOPT should already be getting built and tested. It's installed through conda and the proper environment variable is set. I admit I didn't check to make sure the tests for IPOPT were getting run. Are you sure they're not getting run?

jackm97: Sorry, I realized for IPOPT you had no way of knowing cause the tests weren't shown. The requested flags were added and IPOPT is indeed in there

nwu63: Ah right I forgot that IPOPT is in the conda YAML file already. Exactly, now we can check and make sure they're getting tested. Unfortunately in the current implementation if the imports fail we just skip those optimizers, which may mask actual installation issues.

nwu63:
Okay the only unusual item in the log is the following:
```
    about to run test_rosenbrock.py:TestRosenbrock.test_optimization_0_IPOPT .\test_rosenbrock.py:TestRosenbrock.test_optimization_0_IPOPT ... SKIP (00:00:0.00, 0 MB)
  File "C:\Miniconda\envs\pyos-build\lib\site-packages\testflo\test.py", line 418, in _try_call
    func()
  File "D:\a\pyoptsparse\pyoptsparse\tests\testing_utils.py", line 459, in tearDown
    for _, suffix in OUTPUT_FILENAMES[self.optName].items():
AttributeError: 'TestRosenbrock' object has no attribute 'optName'
```

This test should not be skipped, but the AttributeError should also raise a test failure instead of skipping it, so I'm not sure what's going on. I'm going to be extremely busy in the next 2 weeks so if others don't get to it first, I will take a look when I'm back.

jackm97:
> Okay the only unusual item in the log is the following:
> 
> ```
>     about to run test_rosenbrock.py:TestRosenbrock.test_optimization_0_IPOPT .\test_rosenbrock.py:TestRosenbrock.test_optimization_0_IPOPT ... SKIP (00:00:0.00, 0 MB)
> Traceback (most recent call last):
>   File "C:\Miniconda\envs\pyos-build\lib\site-packages\testflo\test.py", line 418, in _try_call
>     func()
>   File "D:\a\pyoptsparse\pyoptsparse\tests\testing_utils.py", line 459, in tearDown
>     for _, suffix in OUTPUT_FILENAMES[self.optName].items():
> AttributeError: 'TestRosenbrock' object has no attribute 'optName'
> ```
> 
> This test should not be skipped, but the AttributeError should also raise a test failure instead of skipping it, so I'm not sure what's going on. I'm going to be extremely busy in the next 2 weeks so if others don't get to it first, I will take a look when I'm back.

I believe this test actually should be skipped. It was one of the ones that failed on Windows for unknown reasons. For some reason it converges to `92` but the test expects converging to `0`, if I'm not mistaken. The weirdest part about the test is that I have gotten it to pass in the past but I'm unsure what exactly made it pass. My best guess is a different `IPOPT` version. It was included in my original "Windows is working with a disclaimer" comment. I hope I can poke around at some point and either identify the problem or at the very least submit a detailed issue.

nwu63:
> I believe this test actually should be skipped. It was one of the ones that failed on Windows for unknown reasons. For some reason it converges to `92` but the test expects converging to `0`, if I'm not mistaken. The weirdest part about the test is that I have gotten it to pass in the past but I'm unsure what exactly made it pass. My best guess is a different `IPOPT` version. It was included in my original "Windows is working with a disclaimer" comment. I hope I can poke around at some point and either identify the problem or at the very least submit a detailed issue.
> 
> Edit: Although this may bring up some doubt in the `IPOPT` implementation on Windows, I will say all our tests that use the `IPOPT` solver are consistent among Linux, MacOS and Windows. So I really don't know what about this test in particular causes it to fail on Windows.

That seems highly suspicious, as the Rosenbrock problem is unequivocally unimodal with the final objective value at 0. Aside from that, I guess there's a minor bug in the skipTest there (there should be no AttributeError) but we can fix that pretty quickly.

jackm97:
> That seems highly suspicious, as the Rosenbrock problem is unequivocally unimodal with the final objective value at 0. Aside from that, I guess there's a minor bug in the skipTest there (there should be no AttributeError) but we can fix that pretty quickly.

@nwu63 Agreed, especially since every other test case works, including our internal tests which are highly nonlinear and potentially ill formed with discontinuities. Is there any chance this could be addressed in another PR? I would like to probe it and work on it with you, since I don't even know what could cause that. If you have any ideas for a quick probe and fix, I'd be willing to give it a shot. But this PR has been open for quite some time now and it would be nice to close and move the bug to a new issue.

What do you think?

marcomangano:
Hey there, I am sorry I haven't been active here for a while, but I have been keeping an eye on the progress.
I will test/take a look at the build process early next week but given the effort you put in testing it I am confident everything is pretty much ready.

Regarding the failed test, I also feel it is a pretty weird behavior. I do not have a deep understanding of the internal cogs of testflo, but it looks the test fails when testing the `tearDown` function in our test script because somehow the class is not initialized correctly and `self.optName` is not defined (yet). This might also explain why we get a skiptTest instead of a failure. In case we want to debug a bit more, we can try and change the order of the tests [here](https://github.com/mdolab/pyoptsparse/blob/4ae37e62dd71311dabf693132b970925677066b8/tests/test_rosenbrock.py#L151) and see if the issue is IPOPT specific or it just fails on the first call of `parameterized`.

I don't think that this test passing is 100% necessary to have this PR merged but we might want to try a couple of things before giving up.

nwu63:
Sorry, just got back from a week-long conference. I'm happy to address the failed tests in a separate PR and work together for further investigations. The way I see it, then there are two things left before we can merge this:

1. Get @marcomangano and maybe @sseraj to test out this PR and optionally review it (it would be good if I'm not the only maintainer that understands what's going on with meson)
2. Fix the build failure on conda-forge (or was that a different issue that we can ignore? The logs are no longer available so I don't remember why they failed)

marcomangano:
>     1. Get @marcomangano and maybe @sseraj to test out this PR and optionally review it (it would be good if I'm not the only maintainer that understands what's going on with meson)

I will test it by Thursday EoD

pyoptsparse/meson.build
```diff
@@ -0,0 +1,104 @@
...
+#    'pyOpt_utils.py',
+#    'pyOpt_variable.py',
+#    'types.py'
+#]
+
+#py3_target.install_sources(
+#  python_sources,
+#  pure: true,
+#  subdir: 'pyoptsparse'
+#)
```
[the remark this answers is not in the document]
nwu63: These are kept because in the future, meson may become pep517 compliant, which means we don't need to invoke pip at all. In such cases, meson would need to know which source files (e.g. all python files) to install for a user, that's what these lines do.

pyoptsparse/meson.build
```diff
@@ -0,0 +1,104 @@
...
+subdir('pySNOPT')
+subdir('pyIPOPT')
+subdir('pySLSQP')
+subdir('pyCONMIN')
+subdir('pyNLPQLP')
+subdir('pyNSGA2')
+subdir('pyPSQP')
+#subdir('pyALPSO')
+#subdir('pyParOpt')
+#subdir('postprocessing')
```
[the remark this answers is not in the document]
nwu63: This is the same as above, in the future we will hopefully use these.

pyoptsparse/meson.build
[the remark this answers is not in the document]
nwu63: I'll comment them out for now, I do think this is helpful for debugging installation issues than manually trying to import these modules.

pyoptsparse/pySNOPT/meson.build
```diff
@@ -0,0 +1,43 @@
...
+  snopt_source = custom_target('snoptmodule.c',
+    input : ['source/f2py/snopt.pyf'],
+    output : ['snoptmodule.c'],
+    command: [py3_command, '-m', 'numpy.f2py', '@INPUT@',
+    '--lower', '--build-dir', 'pyoptsparse/pySNOPT']
+  )
+
+  py3_target.extension_module('snopt',
+    snopt_source,
+    fortranobject_c,
```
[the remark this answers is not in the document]
nwu63: Fixed

pyoptsparse/meson.build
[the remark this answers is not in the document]
marcomangano: Gotcha, I did not realize that the pep517 issue involved these as well. I would just add a `#TODO` in case it's not there already

nwu63:
A couple more things came up as I work with Marco to figure some stuff out:
- currently, there are a few source files (for SNOPT specifically) which contain non-ASCII characters. They're in the comments and compile file but meson seems to dislike that. When I tried this before I just removed them manually, but that's not really a good approach. Is there a way to bypass this limitation in meson? I would prefer not writing a custom shell/python script that strips characters if possible. I guess I'll open an issue on the meson repo and see what they say.
- debugging the build is not super straightforward with the current setup. Despite build failures, pip reports that everything worked fine, and even the meson logs don't actually show the error unless I'm looking at the wrong thing. In order to see the error message, I had to manually create the builddir and call `meson build`. I wonder if there's a way to make this a little bit easier for end users
- one thing I hadn't thought of was to also test this on HPC systems, I think that has to be added to the checklist. I will keep [this comment](https://github.com/mdolab/pyoptsparse/pull/300#issuecomment-1174391620) updated as a checklist as we go through this process.

marcomangano:
>     * currently, there are a few source files (for SNOPT specifically) which contain non-ASCII characters. They're in the comments and compile fine but meson seems to dislike that. When I tried this before I just removed them manually, but that's not really a good approach. I've opened [Stripping non-ASCII characters mesonbuild/meson#10571](https://github.com/mesonbuild/meson/issues/10571) and they've already opened a PR in response. Woot!

That's impressive, kudos to you and the Meson team! I assume we'll need to add a requirement for Meson>=0.63.1 whenever that's out, it looks like their release rate is pretty fast paced so probably that will be addressed before we complete all the testing on our side.

>     * debugging the build is not super straightforward with the current setup. Despite build failures, pip reports that everything worked fine, and even the meson logs don't actually show the error unless I'm looking at the wrong thing. In order to see the error message, I had to manually create the builddir and call `meson compile`. I wonder if there's a way to make this a little bit easier for end users

Yeah, I like the clarity of the scripts but it really took me a bit just to figure out the issues I posted yesterday. We might add some debugging tips to the docs to help users in the short term.

>     * one thing I hadn't thought of was to also test this on HPC systems, I think that has to be added to the checklist. I will keep [this comment](https://github.com/mdolab/pyoptsparse/pull/300#issuecomment-1174391620) updated as a checklist as we go through this process.

This is another crucial point, we should talk offline about this.

> @marcomangano I believe I fixed your issues. SNOPT didn't compile correctly due to encoding problems and we can talk offline on how I fixed it. The directory that was left behind is because the build actually didn't finish even though pip reported success. This is exactly what I'm talking about above, we should figure out a better way to detect failed builds.

Copy that.

jackm97:
I see some back and forth has picked up. I'm happy to see it! And really awesome that meson was so quick to address the ascii issue. Unfortunately this month is crazy for me as Im moving and have multiple deadlines for work and research. 

That said, I am following along and if you tag me for something I will do my best to get to it as quick as a I can. 

Our team will be exploring SNOPT soon, will the docs be updated with the changes for compiling SNOPT?

Tangentially related, there is a PEP517 Meson backend [here](https://thiblahute.gitlab.io/mesonpep517/). I've seen good things floated around if I remember correctly. I might take a stab at it after this PR! But the weird test failures on Windows are probably higher priority and I'd like to help with that if I can.

pyoptsparse/pySNOPT/meson.build
```diff
@@ -0,0 +1,43 @@
...
+  message('SNOPT source code found!')
+  HAS_SNOPT = true
+else
+  message('SNOPT not found!')
+  HAS_SNOPT = false
+endif
+
+if HAS_SNOPT
+  c = run_command('source' / 'grab-all-fortran-files.py', check: true)
+  snopt_source_files = c.stdout().strip().split('\n')
```
sseraj, line 12: Can we do this for other optimizers as well? Listing out all the files does not seem like the best approach

sseraj (commented):
I checked that this works on my workstation and on NAS. The build on NAS picked up gfortran the same way the distutils build does. I'm not sure how to specify ifort instead.

nwu63:
@jackm97 no worries, I think we can sort out most of these, we'll ping you if we need anything. The SNOPT changes just involved updating the encoding of a few source files from latin1 to UTF8, we can discuss offline if you run into problems.

> I'm not sure how to specify ifort instead.

I believe this is done similar to CMake, via the `FC` environment variable. See [here](https://mesonbuild.com/Reference-tables.html#compiler-and-linker-flag-environment-variables)

pyoptsparse/pySNOPT/meson.build
nwu63: So this type of approach is discouraged in meson, as explained [here](https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard). I personally think the other optimizers don't have quite enough files to need this, especially since we are providing all the source files in pyOptSparse, so there is little advantage. Plus, we are not likely to be adding new files or anything that would impact this.

meson.build
```diff
@@ -0,0 +1,43 @@
...
+# Here we differentiate from the python used by meson, py3_command, and that python target, py3_target. This is useful
+# when cross compiling like on conda-forge
+py_mod = import('python')
+py3_command = py_mod.find_installation()
+if get_option('python_target') != ''
+    py3_target = py_mod.find_installation(get_option('python_target'))
+else
+    py3_target = py3_command
+endif
+py3_dep = py3_target.dependency()
```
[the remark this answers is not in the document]
nwu63: No, this is essentially some metadata for the python installation. It is used when building extension modules, similar to (but I don't think exactly matches) how we use `python-config` in other repos.

setup.py
```diff
@@ -42,12 +30,24 @@ def run_meson_build():
     )
     sysargs = meson_call.split(" ")
     sysargs = [arg for arg in sysargs if arg != ""]
-    meson_main(sysargs)
+    p1 = subprocess.run(sysargs, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    setup_log = os.path.join(staging_dir, "setup.log")
+    with open(setup_log, "wb") as f:
+        f.write(p1.stdout)
```
marcomangano, line 36: Nice way to cleanup the terminal output and store the logs!
I assume writing the stdout after the process is completed is the standard approach within `subprocess`, but I wonder if there are corner cases where a major failure can lead to the logs to be not written out

nwu63: Yep it should spit out the log before throwing any errors. I did test this locally and it seems to work.

nwu63:
Hey, so there are two main remaining TODOs (see [this comment](https://github.com/mdolab/pyoptsparse/pull/300#issuecomment-1174391620)). I think this PR is ready so @sseraj and @marcomangano please review this PR and approve as you see fit. @jackm97 do you think you can figure out the conda build? Or do you think that's not necessary for this to get merged?

I guess we also want to bump the minor version? I can do that shortly.

jackm97:
> Hey, so there are two main remaining TODOs (see [this comment](https://github.com/mdolab/pyoptsparse/pull/300#issuecomment-1174391620)). I think this PR is ready so @sseraj and @marcomangano please review this PR and approve as you see fit. @jackm97 do you think you can figure out the conda build? Or do you think that's not necessary for this to get merged?
> 
> I guess we also want to bump the minor version? I can do that shortly.

@nwu63  Yeah I can look at the conda-forge recipe. It should just be a matter of changing the source url. The build failure that happened was fixed I believe at some point but I'll get back into it and make sure. There are two approaches I see:

1. wait until the PR is merged into main, and then I can change to source url to main and version bump the conda-recipe
2. get it running now with the PR url, do a version bump, and then once it's merged to main change the url back to main.

Either way, the conda-forge PR should probably merged after the merge to main right?

nwu63:
> @nwu63 Yeah I can look at the conda-forge recipe. It should just be a matter of changing the source url. The build failure that happened was fixed I believe at some point but I'll get back into it and make sure. There are two approaches I see:
> 
>   1. wait until the PR is merged into main, and then I can change to source url to main and version bump the conda-recipe
> 
>   2. get it running now with the PR url, do a version bump, and then once it's merged to main change the url back to main.
> 
> 
> Either way, the conda-forge PR should probably merged after the merge to main right?


Yeah this PR should be merged first, I was just wondering if we want to test this stuff in conda-forge before it's merged, in case we need to change something here for conda to build.

sseraj:
I checked that the build works with Intel compilers on NAS. The new logging with subprocess is also nicer.
I'm not sure why one IPOPT test is failing on the Windows job though.

jackm97:
> Yeah this PR should be merged first, I was just wondering if we want to test this stuff in conda-forge before it's merged, in case we need to change something here for conda to build.

That makes sense. I'll get to it over the weekend or early next week! Looking forward to a Windows build on conda

sseraj (approved):

jackm97:
Hi all, some of the Windows tests failed on the conda-forge build. It's file permission errors which I find to be a common theme. On unix systems, as I'm sure you know, trying to remove a file that doesn't exist doesn't fail but it does on Windows. Could we add checks in the tests if files exist and are not being used by another process before read/write/delete? This is a problem within our own test framework as well. 

I could do this but I'm still not that familiar with pyoptpsparse tests so if someone else would like to take it up I'll gladly pass the baton haha

But besides that, the builds on conda-forge all go to completion and tests succeed (sans minor Windows file errors and the disabled Windows tests)

nwu63:
@jackm97 I've pushed up a fix, please try it out. If it works and we get the Azure job to pass, then we can safely merge this one (and then update the upstream on conda before merging that PR).

jackm97:
@nwu63 All builds passed!

nwu63 (approved):
LGTM! Merging :tada:

PR #305, opened 2022-09-27
Enable user requested termination if IPOPT

swryan:
## Purpose

The intent of this change is to enable the IPOPT Optimizer to respond to a user requested termination in the same way that the SNOPT Optimizer does.

As with SNOPT, a value of 2 returned in the `fail` flag from a user evaluation function will be interpreted as a request to terminate the optimization.  This is accomplished by means of the `intermediate_callback` in the IPOPT API.

## Expected time until merged

Not urgent.

## Type of change

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

The existing `user_termination` test for SNOPT has been updated to test IPOPT as well.

## Checklist

- [X] I have run `flake8` and `black` to make sure the Python code adheres to PEP-8 and is consistently formatted
  - I ran `black` and `flake8` but left a few existing lines that `flake8` complained were over 80 columns
  - my `black` made some changes that your `black` didn't like
- [ ] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable
- [X] I have run unit and regression tests which pass locally with my changes
- [X] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #310, opened 2022-10-18
Hard code meson version for Windows build so that tests pass

jackm97:
#305 exposed an issue with using the latest Meson to build Windows. Turns out, the conda-forge feedstock already hardcoded the Meson version to 0.61 which would explain why there haven't been issues with the conda-forge build.

To get the tests to pass, I hard coded the Meson version in the `.github/environment.yaml` file to 0.61. Apparently, an issue has been opened in the Meson repo [here](https://github.com/mesonbuild/meson/issues/10778) discussing the issue. Until this is fixed, it seems 0.61 is the way to go.

### DIFF
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -4,7 +4,7 @@ dependencies:
   - numpy >=1.16
   - ipopt
   - swig
-  - meson >=0.60
+  - meson =0.61
   - compilers
   - pkg-config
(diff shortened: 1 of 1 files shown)


sseraj (approved):

marcomangano (approved):
Thanks for the fix!

PR #312, opened 2022-11-07
Updated IPOPT build instructions for other linear solvers

eytanadler:
## Purpose
The current docs have a note about building IPOPT for pyOptSparse with linear solvers other than mumps. Now that meson is used, this is outdated.

## Type of change

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -6,5 +6,5 @@ import_heading_stdlib=Standard Python modules
 import_heading_thirdparty=External modules
 import_heading_firstparty=First party modules
 import_heading_localfolder=Local modules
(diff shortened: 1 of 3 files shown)


nwu63:
We should probably clean up some more stuff. There are multiple references to `setup.py` (see [here](https://github.com/mdolab/pyoptsparse/search?q=setup.py))but only the root one remains (and should not be modified by the user). Let me know if it's not clear what needs to be removed/updated.

nwu63 (approved):
LGTM

PR #330, opened 2023-02-25
Fix `sol.fStar` for NSGA2

nwu63:
## Purpose

This PR fixes the `sol.fStar` value for NSGA2. Basically, the optimized f value was not being passed to `createSolution`.

## Expected time until merged

2 days

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
An associated test is added which fails without the fix.

## Checklist

### DIFF
--- a/pyoptsparse/pyNSGA2/pyNSGA2.py
+++ b/pyoptsparse/pyNSGA2/pyNSGA2.py
@@ -188,8 +188,9 @@ class NSGA2(Optimizer):
             for i in range(n):
                 xstar[i] = nsga2.doubleArray_getitem(x, i)
 
-            for i in range(len_ff):
-                ff[i] = nsga2.doubleArray_getitem(f, i)
+            if len_ff > 1:
+                for i in range(len_ff):
+                    ff[i] = nsga2.doubleArray_getitem(f, i)
 
             # Create the optimization solution
             sol = self._createSolution(optTime, sol_inform, ff, xstar)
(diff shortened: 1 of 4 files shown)


marcomangano (approved):
Discussed offline, LGTM

marcomangano (dismissed):

pyoptsparse/pyNSGA2/pyNSGA2.py
```diff
@@ -188,6 +188,12 @@ def objconfunc(nreal, nobj, ncon, x, f, g):
             for i in range(n):
                 xstar[i] = nsga2.doubleArray_getitem(x, i)
 
+            if len_ff > 1:
+                for i in range(len_ff):
+                    ff[i] = nsga2.doubleArray_getitem(f, i)
```
sseraj, line 193: Can you add some comments to explain what's happening here? I don't understand why `ff` has to be overwritten.

marcomangano: If I am not mistaken, `ff` is just initialized (as a 0-length array) in `_assembleObjective`. Before this fix, `fStar` was never updated with the value of the optimal objective `f`, returned from the actual optimizer called by `nsga2.nsga2()`. 
Moreover, f is a swig object  so `nsga2.doubleArray_getitem` is necessary to convert it to a numpy array.

marcomangano (approved):

PR #333, opened 2023-03-09
Update informs list to be consistent with SNOPT 7.7 + updating docstrings for flake8 rst checks

### LINKED ISSUE #332: Missing pySNOPT information code for timeout
## Description

When running SNOPT through `pySNOPT`, using the `Time limit` SNOPT option, `pyoptsparse`/`pySNOPT` fails with a key error if the time limit is reached, because [the code for timeout is not handled in `pySNOPT`](https://github.com/mdolab/pyoptsparse/blob/main/pyoptsparse/pySNOPT/pySNOPT.py#L137).

### Steps to reproduce issue

Run any SNOPT optimization with arg `options= {"Time limit": 20}`.

### Current behavior

`KeyError: 34`

### Expected behavior

Return result, showing [informational code #34 from p. 42 of the SNOPT manual](http://www.sbsi-sol-optimize.com/manuals/SNOPT%20Manual.pdf).

### Suggested fix:

Insert
```
            34: "time limit reached",
```
at [pySNOPT.py, line 137](https://github.com/mdolab/pyoptsparse/blob/main/pyoptsparse/pySNOPT/pySNOPT.py#L137)

marcomangano:
## Purpose

I updated the `informs` dictionary to include all the supported `INFO` exit codes in SNOPT 7.7 (see list on the [manual](http://www.sbsi-sol-optimize.com/manuals/SNOPT%20Manual.pdf), page 42).
A few of the current codes are not used in v7.7 anymore, so I removed them from the list. This will not affect backward compatibility as we don't support versions <7.7.

Closes #332 

P.S. I edited a few docstrings that were creating issues with the new flake8 rst checks, since this is the first PR affected by the new check for this repo

## Expected time until merged

1-2 days

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/doc/advancedFeatures.rst
+++ b/doc/advancedFeatures.rst
@@ -62,6 +62,12 @@ Because the hot start process will store all the previous "restarted" iterations
 
 Time limit (for SNOPT only)
 ---------------------------
+
+.. note::
+
+   Since SNOPT 7.7, the user should rely on the ``Time Limit`` SNOPT option instead of the ``timeLimit`` argument of the ``Optimizer`` instance to set the maximum optimization time.
+
+
 The :ref:`optimizer` class in pyOptSparse has an attribute used to set the maximum allowable wall time for optimizations using SNOPT.
 The code will exit gracefully when such time limit is reached.
 This feature is particularly useful when running a time-constrained job, as in the case of most HPC systems.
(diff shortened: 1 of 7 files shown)


sseraj (commented):
It seems like we have only supported 7.7.x for a while now (since v2.0). Maybe we can just remove the old informs?

marcomangano:
> It seems like we have only supported 7.7.x for a while now (since v2.0). Maybe we can just remove the old informs?

Oh right, I was being overly conservative. Removing now

nwu63:
Have you tried the time limit option in SNOPT? I don't think we should recommend things until we've tried it and feel comfortable with replacing it.

marcomangano:
> Have you tried the time limit option in SNOPT? I don't think we should recommend things until we've tried it and feel comfortable with replacing it.

I tested it and it works as intended, but I will a test for this in the next few days

sseraj (approved):

lamkina (approved):

Issue #337, opened 2023-04-26
Allow `snopt` module to be loaded non-relatively


whophil:
# Description of feature

Currently the `snopt.*.so` binding module is loaded relatively to pySNOPT.py - it is expected to live in the same folder.

https://github.com/mdolab/pyoptsparse/blob/40af7dcad39537f7d6ff26583ad2e9b7422bbcdf/pyoptsparse/pySNOPT/pySNOPT.py#L7

This is a problem when installing `pyoptsparse` from conda-forge, because the `snopt` module is not included in the conda-forge release, due to license restrictions.

Consider the case where a user wishes to:
1. Install `pyoptsparse` from conda-forge
2. Provide their own pre-compiled `snopt.*.so` binding module

As currently implemented, the user would have to copy it into the environment's site-packages folder to allow `snopt` to be imported correctly. This is not desirable.

# Potential solution

An environment variable which allows the `snopt` module to be loaded _non-relatively_ should do the trick, as a user can put whatever path onto their module search path with `PYTHONPATH`.

whophil:
Closed by #338

PR #338, opened 2023-04-26
Add ability to specify search path for SNOPT import with env var

whophil:
## Purpose

Addresses #337 

I am totally open to suggestions on the name of this environment variable. For now it is explicit as possible.

## Expected time until merged

Quickly - due to small scope of change, not urgency.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/doc/optimizers/SNOPT.rst
+++ b/doc/optimizers/SNOPT.rst
@@ -32,13 +32,12 @@ If you are installing via conda and would like to use SNOPT, you will need to bu
 
 Suppose you have built the binding file, producing ``snopt.cpython-310.so``, living in the folder ``~/snopt-bind``.
 
-To use this module, set the environment variable, `PYOPTSPARSE_SNOPT_PATH_PREPEND`, e.g.:
+To use this module, set the environment variable, `PYOPTSPARSE_IMPORT_SNOPT_FROM`, e.g.:
 ```
 PYOPTSPARSE_IMPORT_SNOPT_FROM=~/snopt-bind/
 ```
 
-This will attempt to load the ``snopt`` binding module from ``~/snopt-bind``. If the module cannot be loaded from this
-path, a warning will be raised at import time, and an error will be raised if attempting to run the SNOPT optimizer.
+This will attempt to load the ``snopt`` binding module from ``~/snopt-bind``. If the module cannot be loaded from this path, a warning will be raised at import time, and an error will be raised if attempting to run the SNOPT optimizer.
 
 Options
 -------
--- a/doc/optimizers/SNOPT.rst
+++ b/doc/optimizers/SNOPT.rst
@@ -13,6 +13,9 @@ BFGS quasi-Newton update.
 
 Installation
 ------------
+
+Building from source
+********************
 SNOPT is available for purchase `here 
 <http://www.sbsi-sol-optimize.com/asp/sol_snopt.htm>`_. Upon purchase, you should receive a zip file. Within the zip file, there is a folder called ``src``. To use SNOPT with pyoptsparse, paste all files from ``src`` except snopth.f into ``pyoptsparse/pySNOPT/source``.
 
@@ -20,6 +23,17 @@ From v2.0 onwards, only SNOPT v7.7.x is officially supported.
 To use pyOptSparse with previous versions of SNOPT, please checkout release v1.2.
 We currently test v7.7.7 and v7.7.1.
 
+Installation by conda
+*********************
+When installing via conda, all pyoptsparse binaries are pre-compiled and installed as part of the package. However, the `snopt` binding module cannot be included as part of the package due to license restrictions.
+
+If you are installing via conda and would like to use SNOPT, you will need to build the `snopt` binding module on your own and inform `pyoptsparse` that it should use that library.
+
+Suppose you have built the binding file, producing `snopt.cpython-310.so`, living in the folder `~/snopt-bind`.
+
+To inform `pyoptsparse` that it should use this module, set the following environment variables:
+1. `PYOPTSPARSE_IMPORT_SNOPT_ABSOLUTE=1` to inform pyoptsparse that it should import `snopt` bindings absolute, rather than relative (the default)
+2. `PYTHONPATH=~/snopt-bind:${PYTHONPATH}` to put the `~/snopt-bind` library on the module search path.
 
 Options
 -------
--- a/doc/optimizers/SNOPT.rst
+++ b/doc/optimizers/SNOPT.rst
@@ -34,12 +34,11 @@ Suppose you have built the binding file, producing ``snopt.cpython-310.so``, liv
 
 To use this module, set the environment variable, `PYOPTSPARSE_SNOPT_PATH_PREPEND`, e.g.:
 ```
-PYOPTSPARSE_SNOPT_PATH_PREPEND=~/snopt-bind/
+PYOPTSPARSE_IMPORT_SNOPT_FROM=~/snopt-bind/
 ```
 
-Setting this variable does the following:
-1. Temporarily prepends the given path to the `PYTHONPATH` during import of `snopt`
-2. Informs pyOptSparse that it should attempt to load `snopt` _absolutely_ (by searching the `PYTHONPATH`) before attempting to import relatively (the default behavior if `PYOPTSPARSE_SNOPT_PATH_PREPEND` is not set).
+This will attempt to load the ``snopt`` binding module from ``~/snopt-bind``. If the module cannot be loaded from this
+path, a warning will be raised at import time, and an error will be raised if attempting to run the SNOPT optimizer.
 
 Options
 -------
--- a/doc/optimizers/SNOPT.rst
+++ b/doc/optimizers/SNOPT.rst
@@ -32,10 +32,11 @@ If you are installing via conda and would like to use SNOPT, you will need to bu
 
 Suppose you have built the binding file, producing ``snopt.cpython-310.so``, living in the folder ``~/snopt-bind``.
 
-To use this module, set the environment variable, `PYOPTSPARSE_IMPORT_SNOPT_FROM`, e.g.:
-```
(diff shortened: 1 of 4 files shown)


marcomangano (commented):
Thanks for the PR Phil! I just have two quick comments:
- I like using a self-explanatory name for the environmental variable, even if long. I am just not convinced about the `_NONRELATIVE` part, it might not be intuitive for users. What about something along the lines of `PYOPTSPARSE_LOAD_COMPILED_SNOPT`? Would that be ambiguous? I have no strong opinions on this name myself .
- Could you update the docs [here](https://github.com/whophil/pyoptsparse/blob/feature/snopt-module-load-options/doc/optimizers/SNOPT.rst#installation)? Just clarifying the scenario in which this approach is recommended (as you did in #337) and a snippet with the updated env variables  is enough imho

whophil:
@marcomangano pushed a change renaming the variable to `PYOPTSPARSE_IMPORT_SNOPT_ABSOLUTE`, and also added info to docs.

doc/optimizers/SNOPT.rst
```diff
@@ -13,13 +13,27 @@ BFGS quasi-Newton update.
...
 To use pyOptSparse with previous versions of SNOPT, please checkout release v1.2.
 We currently test v7.7.7 and v7.7.1.
 
+Installation by conda
+*********************
+When installing via conda, all pyoptsparse binaries are pre-compiled and installed as part of the package. However, the `snopt` binding module cannot be included as part of the package due to license restrictions.
+
+If you are installing via conda and would like to use SNOPT, you will need to build the `snopt` binding module on your own and inform `pyoptsparse` that it should use that library.
+
+Suppose you have built the binding file, producing `snopt.cpython-310.so`, living in the folder `~/snopt-bind`.
```
eirikurj, line 32: We generally use `` (double backticks) to display paths and variables etc., as single will make it italic in RST. I suggest updating where appropriate.

doc/optimizers/SNOPT.rst
```diff
@@ -13,13 +13,27 @@ BFGS quasi-Newton update.
...
+
+To inform `pyoptsparse` that it should use this module, set the following environment variables:
+1. `PYOPTSPARSE_IMPORT_SNOPT_ABSOLUTE=1` to inform pyoptsparse that it should import `snopt` bindings absolute, rather than relative (the default)
+2. `PYTHONPATH=~/snopt-bind:${PYTHONPATH}` to put the `~/snopt-bind` library on the module search path.
```
eirikurj, line 36: Just a thought, but is there a need for a separate flag variable? A single env variable could be used to hold the path directly to the library and control the import. It could be named `PYOPTSPARSE_SNOPT_IMPORT_PATH` or `PYOPTSPARSE_SNOPT_LIB_PATH` (or similar). Then in the if set check in `pySNOPT.py` it could be directly added to the path before loaded. This would be a more local approach, as `PYTHONPATH` applies to everything.

whophil: Implemented it this way now, updated docs appropriately.

whophil:
The failing test seems to be related to NSGA segfaults (#340), otherwise this is ready.

whophil:
A clarification on how this behaves - 

The original search order for `snopt` was:
1. The directory relative to `pySNOPT.py`

With this change, if you specify a path with `PYOPTSPARSE_SNOPT_PATH_PREPEND`, the search order for `snopt` becomes:
1. The path specified by `PYOPTSPARSE_SNOPT_PATH_PREPEND`
2. Standard `PYTHONPATH`
3. The directory relative to `pySNOPT.py`

The second item may be unexpected, so I wanted to point it out.

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -26,6 +22,38 @@
...
+    else:
+        from . import snopt  # isort: skip
+    return snopt
+
+
+_SNOPT_PATH_PREPEND = os.environ.get("PYOPTSPARSE_SNOPT_PATH_PREPEND", None)
+if _SNOPT_PATH_PREPEND is not None:
+    try:
+        snopt = _import_snopt(absolute=True, path_prepend=_SNOPT_PATH_PREPEND)
+    except ImportError:
```
marcomangano, line 44: I think it could be helpful to add here a warning message to make the user aware that `snopt` could not be imported from `PYTHONPATH`  - most likely `PYOPTSPARSE_SNOPT_PATH_PREPEND` is incorrectly set.

marcomangano (commented):
Thanks for the update. Aside the comment below, I feel your note on the search order should be included in the documentation.

I see how searching through the "original" `PYTHONPATH` is inevitable with this approach. I don't have anything against it, but I think we need some kind of note and/or warning. I can see how more or less intentionally a user could set an invalid env variable but still end up importing any `snopt` library already existing on the path. I don't think the logic should change though. Do you have other thoughts @eirikurj

whophil:
@marcomangano there is one other option, which has no hidden behavior. I will explain it in words and wait for your feedback before implementing.

1. If the env var `PYOPTSPARSE_LOAD_SNOPT_FROM` exists, prepend `PYTHONPATH` with that path and `import snopt`.
2. Check where `snopt` was imported from.
3. If it wasn't imported from the specified path, don't use it! Set `snopt = None`

There is also the option to **not** attempt the relative import as a fallback if the absolute import fails. This feels a little more intuitive to me.

Personally, I would prefer an **error** to be raised if `PYOPTSPARSE_LOAD_SNOPT_FROM` is specified and `snopt` cannot be successfully loaded from that path, but that seems to be out of line with the `snopt = None` approach.

marcomangano:
>     1. If the env var `PYOPTSPARSE_LOAD_SNOPT_FROM` exists, prepend `PYTHONPATH` with that path and `import snopt`.
> 
>     2. Check where `snopt` was imported from.
> 
>     3. If it wasn't imported from the specified path, don't use it! Set `snopt = None`

I like this approach, it avoids the ambiguous scenarios I mentioned above.

> There is also the option to **not** attempt the relative import as a fallback if the absolute import fails. This feels a little more intuitive to me.

I also agree with this. pyOptSparse is either importing the library specified in the path or dropping the import, no ambiguous fallbacks.

> Personally, I would prefer an **error** to be raised if `PYOPTSPARSE_LOAD_SNOPT_FROM` is specified and `snopt` cannot be successfully loaded from that path, but that seems to be out of line with the `snopt = None` approach.

I think this might be redundant, as we indeed have a check on the correct import [here](https://github.com/whophil/pyoptsparse/blob/8354b4a2abd59696d1af06cd125b804d311668c5/pyoptsparse/pySNOPT/pySNOPT.py#L89-L93). Maybe we can print a warning for this scenario, but if we end up implementing your latest suggestion I think there is no need for an additional error imho.

marcomangano (approved):
This looks good to me! Just fixed a minor inconsistency in the docs. Maybe @eirikurj has some final comments on this?

whophil:
> This looks good to me! Just fixed a minor inconsistency in the docs. Maybe @eirikurj has some final comments on this?

Thanks @marcomangano !

I just pushed one other basic test for `_import_snopt_from_path`. It doesn't complete the test coverage, but it's better than nothing.

doc/optimizers/SNOPT.rst
```diff
@@ -13,13 +13,31 @@ BFGS quasi-Newton update.
...
+However, the `snopt` binding module cannot be included as part of the package due to license restrictions.
+
+If you are installing via conda and would like to use SNOPT, you will need to build the `snopt` binding module on your own, and inform `pyoptsparse` that it should use that library.
+
+Suppose you have built the binding file, producing ``snopt.cpython-310.so``, living in the folder ``~/snopt-bind``.
+
+To use this module, set the environment variable, `PYOPTSPARSE_IMPORT_SNOPT_FROM`, e.g.:
+```
+PYOPTSPARSE_IMPORT_SNOPT_FROM=~/snopt-bind/
+```
```
eirikurj, line 38: Did you mean the env to be italic? I suggest we update var with ``. Also for the following code block, in RST please use the `.. code-block:: bash` directive, not the ``` markdown block. It will not be rendered properly.

tests/test_other.py
```diff
@@ -0,0 +1,16 @@
...
+import sys
+import unittest
+
+# First party modules
+from pyoptsparse.pySNOPT.pySNOPT import _import_snopt_from_path
+
+
+class TestImportSnoptFromPath(unittest.TestCase):
+    def test_nonexistent_path(self):
+        assert _import_snopt_from_path("/a/nonexistent/path") is None
```
eirikurj, line 11: Maybe directly use `self.assertIsNone` instead.
Also, do we want to test that the warning is emitted here? Should be able to catch by something like `with self.assertWarns(Warning):` (same below?)

tests/test_other.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def test_sys_path_unchanged(self):
+        path = tuple(sys.path)
+        _import_snopt_from_path("/some/path")
+        assert tuple(sys.path) == path
```
eirikurj, line 16: Maybe directly use `self.assertEqual` instead.

eirikurj (changes requested):
Thanks for your effort on this and the tests. I just have few very minor questions/comments, otherwise it looks good.

tests/test_other.py
whophil: Done, also changed the warning to be `ImportWarning` which seems appropriate.

tests/test_other.py
whophil: Done

doc/optimizers/SNOPT.rst
whophil: Fixed

eirikurj (approved):
Changes look good. Not sure, but the U22 image failed a couple of times installing the build deps. The current failure is the #340 issue, which is unrelated, so I am approving.

marcomangano (approved):

PR #344, opened 2023-05-12
Remove callables from optProb only when serializing in History file

nwu63:
## Purpose

So, let's unpack the story of what's going on.
1. Originally we removed the `objFun` field when serializing optProb since it's a callable
2. When using multiprocessing, it is important to hold onto those references (see #264), so we keep them if possible (see #266)
3. We also added `optProb.sens` in #212, it is also a callable similar to `objFun` but did not receive the same treatment

Now, instead of just doing the same thing as before, I changed the approach of serialization. When we work with the `Optimization` object, we only get rid of the MPI communicator. When we need to serialize it for the History object, we then remove those fields so they are not serialized. This separation of Optimization vs. History is a little bit cleaner IMO.

## Expected time until merged

No rush. Please test this with lab tools (SNOPT + MACH) to make sure I didn't break anything. The CI will also not run SNOPT tests so please test those locally.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

We do not have any tests that use the `optProb.sens` field, but I can add some if maintainers feel this is necessary.

## Checklist

- [x] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable

### DIFF
--- a/pyoptsparse/pyOpt_optimization.py
+++ b/pyoptsparse/pyOpt_optimization.py
@@ -2,7 +2,6 @@
 from collections import OrderedDict
 import copy
 import os
-import pickle
 from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 # External modules
--- a/pyoptsparse/pyOpt_optimization.py
+++ b/pyoptsparse/pyOpt_optimization.py
@@ -1761,15 +1761,7 @@ class Optimization:
         The un-serializable fields are deleted first.
         """
         d = copy.copy(self.__dict__)
-        keysToRemove = ["comm"]
-        try:
-            pickle.dumps(self.objFun)
-        except Exception:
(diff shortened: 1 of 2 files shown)


marcomangano (approved):
LGTM, definitely cleaner
I don't foresee issues but I will try and test this in the next few days

lamkina:
> LGTM, definitely cleaner I don't foresee issues but I will try and test this in the next few days

@marcomangano do you also want to test locally?  If not this is ready to go I think.

marcomangano:
> > LGTM, definitely cleaner I don't foresee issues but I will try and test this in the next few days
> 
> @marcomangano do you also want to test locally? If not this is ready to go I think.

Thanks for checking! merging it now

PR #346, opened 2023-05-12
Change lambdaStar to always be a dictionary

nwu63:
## Purpose

Currently, for optimizers that do not return the Lagrange multipliers, `Solution.lambdaStar` is set to `None`. IMO it is better to set it to an empty dictionary such that the variable type at least matches. I have done this by updating the `_createSolution()` function.

This feature is technically backwards breaking, so happy to discuss this further with maintainers.

## Expected time until merged

Not urgent.

## Type of change

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

- [ ] I have run `flake8` and `black` to make sure the Python code adheres to PEP-8 and is consistently formatted
- [ ] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable
- [ ] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #347, opened 2023-05-19
Save more variables by default in SNOPT history file

nwu63:
## Purpose

I updated the `snSTOP` function to save more variables by default in the Python layer. The option `Save major iteration variables` now defaults to an empty list. The following variables are saved (the user cannot turn these off)
- `isMajor`
- `nMajor`
- `nMinor`
- `step`
- `feasibility`
- `optimality`
- `merit`
- `condZHZ`
- `penalty`

and the following can be added via the option
- `Hessian`
- `slack`
- `lambda`
- `condZHZ`
- `nS`
- `BSwap`
- `maxVi`
- `penalty_vector`

Happy to discuss names/which ones should be saved by default. I think it's possible to have everything in the major iteration log saved, but the defaults I feel are the ones people actually care about (at least given my limited knowledge of SNOPT).

## Expected time until merged

A few days. Maybe we can do a minor release after this and #346 are both merged?

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

I don't believe this is a backwards-breaking change, but happy to discuss. Also please run tests since the CI will not check SNOPT-related stuff.

## Testing

I updated the existing tests. As usual, the CI will not test SNOPT so please test this locally.

## Checklist

### DIFF
--- a/doc/optimizers/SNOPT_options.yaml
+++ b/doc/optimizers/SNOPT_options.yaml
@@ -82,6 +82,17 @@ Save major iteration variables:
     - ``maxVi``
     - ``penalty_vector``
 
+    In addition, a set of default parameters are saved to the history file and cannot be changed. These are
+
+    - ``nMajor``
+    - ``nMinor``
+    - ``step``
+    - ``feasibility``
+    - ``optimality``
+    - ``merit``
+    - ``condZHZ``
+    - ``penalty``
+
 Return work arrays:
   desc: >
     This option is unique to the Python wrapper.
--- a/doc/optimizers/SNOPT_options.yaml
+++ b/doc/optimizers/SNOPT_options.yaml
@@ -77,7 +77,6 @@ Save major iteration variables:
     - ``Hessian``
     - ``slack``
     - ``lambda``
-    - ``condZHZ``
     - ``nS``
     - ``BSwap``
     - ``maxVi``
--- a/doc/optimizers/SNOPT_options.yaml
+++ b/doc/optimizers/SNOPT_options.yaml
@@ -71,9 +71,17 @@ Total real workspace:
 
 Save major iteration variables:
   desc: >
-    This option is unique to the Python wrapper, and takes a list of values which can be saved at each iteration to the History file.
-    This specifies the list of major iteration variables to be stored in the history file.
-    ``Hessian``, ``slack``, ``lambda`` and ``condZHZ`` are also supported.
+    This option is unique to the Python wrapper, and takes a list of values which can be saved at each major iteration to the History file.
+    The possible values are
+
+    - ``Hessian``
+    - ``slack``
+    - ``lambda``
+    - ``condZHZ``
+    - ``nS``
+    - ``BSwap``
+    - ``maxVi``
+    - ``penalty_vector``
 
 Return work arrays:
   desc: >
--- a/pyoptsparse/pySNOPT/pySNOPT.py
+++ b/pyoptsparse/pySNOPT/pySNOPT.py
@@ -666,7 +666,7 @@ class SNOPT(Optimizer):
             elif saveVar == "maxVi":
                 iterDict[saveVar] = maxvi
             else:
-                raise Error(f"Received unkown SNOPT save variable {saveVar}. "
+                raise Error(f"Received unknown SNOPT save variable {saveVar}. "
                             + "Please see 'Save major iteration variables' option in the pyOptSparse documentation "
                             + "under 'SNOPT'.")
         if self.storeHistory:
--- a/pyoptsparse/pySNOPT/pySNOPT.py
+++ b/pyoptsparse/pySNOPT/pySNOPT.py
@@ -614,7 +614,7 @@ class SNOPT(Optimizer):
         H = np.matmul(Umat.T, Umat)
         return H
 
-    def _getPenaltyParam(self, iw, rw):
+    def _getPenaltyVector(self, iw, rw):
         """
         Retrieves the full penalty parameter vector from the work arrays.
         """
@@ -651,7 +651,7 @@ class SNOPT(Optimizer):
         }
         for saveVar in self.getOption("Save major iteration variables"):
             if saveVar == "penalty_vector":
-                iterDict[saveVar] = self._getPenaltyParam(iw, rw),
+                iterDict[saveVar] = self._getPenaltyVector(iw, rw),
             elif saveVar == "Hessian":
                 H = self._getHessian(iw, rw)
                 iterDict[saveVar] = H
@@ -666,7 +666,7 @@ class SNOPT(Optimizer):
             elif saveVar == "maxVi":
                 iterDict[saveVar] = maxvi
             else:
-                warnings.warn(f"Received unknown SNOPT save variable {saveVar}")
+                warnings.warn(f"Received unknown SNOPT save variable {saveVar}", stacklevel=2)
         if self.storeHistory:
             currX = x[:n]  # only the first n component is x, the rest are the slacks
             if nmajor == 0:
--- a/pyoptsparse/pySNOPT/pySNOPT.py
+++ b/pyoptsparse/pySNOPT/pySNOPT.py
@@ -137,10 +137,7 @@ class SNOPT(Optimizer):
             "Total character workspace": [int, None],
             "Total integer workspace": [int, None],
             "Total real workspace": [int, None],
-            "Save major iteration variables": [
-                list,
-                ["step", "merit", "feasibility", "optimality", "penalty"],
-            ],
+            "Save major iteration variables": [list, []],
             "Return work arrays": [bool, False],
             "snSTOP function handle": [(type(None), type(lambda: None)), None],
         }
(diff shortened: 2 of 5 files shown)


tests/test_other.py
```diff
@@ -1,9 +1,14 @@
 # Standard Python modules
+import os
 import sys
 import unittest
 
+# we have to unset this environment variable because otherwise when we import `_import_snopt_from_path`
+# the snopt module gets automatically imported, thus failing the import test below
+os.environ.pop("PYOPTSPARSE_IMPORT_SNOPT_FROM", None)
```
sseraj, line 8: Was this working for the CI because we don't set the environment variable in the Docker images?

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -627,43 +624,49 @@ def _getPenaltyParam(self, iw, rw):
...
             elif saveVar == "lambda":
                 iterDict[saveVar] = pi
+            elif saveVar == "nS":
+                iterDict[saveVar] = ns
+            elif saveVar == "BSwap":
+                iterDict[saveVar] = nswap
+            elif saveVar == "maxVi":
+                iterDict[saveVar] = maxvi
+            else:
+                warnings.warn(f"Received unknown SNOPT save variable {saveVar}")
```
sseraj, line 669: I think an error would be better here

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -627,43 +624,49 @@ def _getPenaltyParam(self, iw, rw):
...
-                iterDict[saveVar] = fmerit
-            elif saveVar == "feasibility":
-                iterDict[saveVar] = primalinf
-            elif saveVar == "optimality":
-                iterDict[saveVar] = dualinf
-            elif saveVar == "penalty":
-                penParam = self._getPenaltyParam(iw, rw)
-                iterDict[saveVar] = penParam
+            if saveVar == "penalty_vector":
+                iterDict[saveVar] = self._getPenaltyParam(iw, rw),
```
sseraj, line 654: Maybe rename `_getPenaltyParam` to `_getPenaltyVector`? Otherwise, I think it's easy to confuse with `penparm`

doc/optimizers/SNOPT_options.yaml
```diff
@@ -71,9 +71,17 @@ Total real workspace:
...
-    This option is unique to the Python wrapper, and takes a list of values which can be saved at each iteration to the History file.
-    This specifies the list of major iteration variables to be stored in the history file.
-    ``Hessian``, ``slack``, ``lambda`` and ``condZHZ`` are also supported.
+    This option is unique to the Python wrapper, and takes a list of values which can be saved at each major iteration to the History file.
+    The possible values are
+
+    - ``Hessian``
+    - ``slack``
+    - ``lambda``
+    - ``condZHZ``
```
sseraj, line 80: This is a default already

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -137,10 +137,7 @@ def _getDefaultOptions() -> Dict[str, Any]:
             "Total character workspace": [int, None],
             "Total integer workspace": [int, None],
             "Total real workspace": [int, None],
-            "Save major iteration variables": [
-                list,
-                ["step", "merit", "feasibility", "optimality", "penalty"],
-            ],
+            "Save major iteration variables": [list, []],
```
sseraj, line 140: I think it would be better to add to the default list than have variables that cannot be turned off. This would make the code easier to understand, and it would be consistent with how we handle saving variables in ADflow.

tests/test_other.py
nwu63: Correct. That being said, the test still fails unless it is run with the `isolated` flag from `testflo`, because it's possible for a previous test to import `SNOPT` first, thereby importing the `snopt` module. I'm not sure of an easy way to fix this if the behaviour we want is to import the `snopt` module if possible at the start of the pySNOPT module.

nwu63: This does not impact CI but will impact people locally who are using this env var.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: I thought about it, and decided on warning since I don't want to kill a whole optimization just because someone made a typo or had a duplicate variable. Can change if preferred.

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: I also debated about this. I think the reason I preferred this is because, if I want to add another one then I have to copy/paste the list of defaults first from somewhere (probably docs) which is a little annoying. Also, given the small storage requirement I don't think it's necessary to allow users to remove the default ones. Happy to change things though.

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -627,43 +624,49 @@ def _getPenaltyParam(self, iw, rw):
...
 
-        returning with iabort != 0 will terminate SNOPT immediately
+        Returns
+        -------
+        iabort : int
+            The return code expected by SNOPT. A non-zero value will terminate SNOPT immediately.
         """
         iterDict = {
             "isMajor": True,
             "nMajor": nmajor,
```
nwu63, line 643: One other thing I want to discuss is the name of these variables. Some match the SNOPT print out but others (e.g. `nMajor`) do not. Do we want to make things consistent in some way? It will break backwards compatibility if your post-processing script is explicitly looking for these names though.

pyoptsparse/pySNOPT/pySNOPT.py
sseraj: What do you think @lamkina?

pyoptsparse/pySNOPT/pySNOPT.py
sseraj: I would be fine with changing these to match SNOPT

pyoptsparse/pySNOPT/pySNOPT.py
sseraj: Thoughts on this @lamkina?

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: Hmm it's a little tricky though, because IMO SNOPT terminology is a little strange:
* `Feasible`/`Optimal` are adjectives being used as nouns
* Major but then Minors (I guess Major is always incremented by at most 1 but still)
* `Itns` will cause confusion with the existing key `iteration` which stores the callCounter so I've opted to skip this one

I guess the only ones I'm unhappy about in the current naming are `nMajor` and `nMinor` vs say `major` and `minor`, the others I'm actually okay with.

nwu63:
Bumping this PR @sseraj @lamkina

pyoptsparse/pySNOPT/pySNOPT.py
nwu63: I'm actually fine with the current names in order to not break backwards compatibility, we can revisit this later if needed.

pyoptsparse/pySNOPT/pySNOPT.py
lamkina: I tend to agree with @nwu63 in this case.  The storage requirement for the default variables is pretty insignificant and from an OptView perspective it's nice to have an expected minimal set of data for every SNOPT case.

From the user perspective, I'm always against extra steps to get simple cases working.  If the proposed alternative means users need to read the docs to understand the default usage of recoring, I think that is too complicated for the intended impact of these changes.

pyoptsparse/pySNOPT/pySNOPT.py
lamkina: I would rather this be an error with a more descriptive message.  If you are expecting to have variables to postprocess afer an optimization, you don't want to have to re-run the entire optimization (especially for high-fidelity cases) just to get that information.  Especially because most users initially won't understand this warning and might just ignore the warning.

An error with a message like `Received unkown SNOPT save variable {saveVar}.  Please see "Save major iteration variables" option in the pyOptSparse documentation under "SNOPT".`

lamkina (commented):
See my comment about changing the warning to an error.  Once that's resolved I think the rest of this is looking good to go.

pyoptsparse/pySNOPT/pySNOPT.py
sseraj: Alright, that fine with me. I would like the defaults to be documented though because the empty list suggests nothing is saved by default.

lamkina: That's a good compromise imo.

pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -627,43 +624,51 @@ def _getPenaltyParam(self, iw, rw):
...
             elif saveVar == "lambda":
                 iterDict[saveVar] = pi
+            elif saveVar == "nS":
+                iterDict[saveVar] = ns
+            elif saveVar == "BSwap":
+                iterDict[saveVar] = nswap
+            elif saveVar == "maxVi":
+                iterDict[saveVar] = maxvi
+            else:
+                raise Error(f"Received unkown SNOPT save variable {saveVar}. "
```
sseraj, line 669: unkown -> unknown

sseraj (approved):

marcomangano (approved):
Sorry for my late review, I agree with the comments above. It's not  really a  new feature and there are no big changes in behavior so I don't think a new release is necessary just for this PR, but we might think about a version bump soon.

PR #348, opened 2023-05-22
More tolerant scalar check

nwu63:
## Purpose

It's possible to have an unsized numpy array, e.g.
```
>>> x = np.array(5)
>>> x.shape
()
```
In such cases, the `isscalar` check fails:
```
>>> np.isscalar(np.array(5))
False
```

In this PR, I refactored this code into a utility function `_broadcast_to_array`, and switched to using `numpy.broadcast_to` which should work for all value types. This simplifies the code somewhat. I have commented below on things that may require attention in the review process.

## Expected time until merged

1-2 days.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I did not add any tests.

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimization.py
+++ b/pyoptsparse/pyOpt_optimization.py
@@ -15,7 +15,18 @@ from .pyOpt_MPI import MPI
 from .pyOpt_constraint import Constraint
 from .pyOpt_error import Error
 from .pyOpt_objective import Objective
-from .pyOpt_utils import ICOL, IDATA, INFINITY, IROW, convertToCOO, convertToCSR, mapToCSR, scaleColumns, scaleRows
+from .pyOpt_utils import (
+    ICOL,
+    IDATA,
+    INFINITY,
+    IROW,
+    convertToCOO,
+    convertToCSR,
+    mapToCSR,
+    scaleColumns,
+    scaleRows,
+    _broadcast_to_array,
+)
 from .pyOpt_variable import Variable
 from .types import Dict1DType, Dict2DType, NumpyType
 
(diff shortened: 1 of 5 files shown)


marcomangano (commented):
This makes sense. I see we have the same code almost exactly replicated (without the casting to real) in the [Constraint class](https://github.com/nwu63/pyoptsparse/blob/4ddff7b72c255b3a1d9868712f4c20d4a75c4336/pyoptsparse/pyOpt_constraint.py#L46-L68) so it might make sense to turn this into a short utility function for consistency. Let me know if you want to do it, I can also work on that but I have little bandwidth in the next few days

pyoptsparse/pyOpt_utils.py
```diff
@@ -529,3 +529,13 @@ def _csc_to_coo(mat: dict) -> dict:
     coo_data = np.array(data)
 
     return {"coo": [coo_rows, coo_cols, coo_data], "shape": mat["shape"]}
+
+
+def _broadcast_to_array(name: str, value, n_vars: int, allow_none: bool = True):
+    try:
+        value = np.broadcast_to(value, (n_vars))
```
nwu63, line 536: Arrays containing None will get broadcast into `dtype=object`. I think this is fine since we only examine individual components of these arrays, e.g. [here](https://github.com/mdolab/pyoptsparse/blob/main/pyoptsparse/pyOpt_constraint.py#L121) for constraints. Design variables are always stored in scalar form so it doesn't matter there.

pyoptsparse/pyOpt_optimization.py
```diff
@@ -229,71 +240,11 @@ def addVarGroup(
...
-                pass
-            else:
-                raise Error(
-                    f"The length of the 'scale' argument to addVarGroup is {len(scale)}, "
-                    + f"but the number of variables in nVars is {nVars}."
-                )
-
-        # ------ Process the offset argument
-        if offset is None:
-            offset = np.ones(nVars)
```
nwu63, line 285: I think this is actually a bug. We should have `offset` default to zero instead of one (see default value above). In any case I don't think we should allow user to pass in `None`.

pyoptsparse/pyOpt_optimization.py
```diff
@@ -229,71 +240,11 @@ def addVarGroup(
...
-                f"The length of the 'value' argument to addVarGroup is {len(value)}, "
-                + f"but the number of variables in nVars is {nVars}."
-            )
-
-        if lower is None:
-            lower = [None for i in range(nVars)]
-        elif np.isscalar(lower):
-            lower = lower * np.ones(nVars)
-        elif len(lower) == nVars:
-            lower = np.atleast_1d(lower).real
```
nwu63, line 249: We throw around `.real` very liberally here, IMO this is bad practice and can mask bugs in the code. It is the user's responsibility to make sure things are passed in with the correct dtype, and we should not just cast things to real here. Note that this is potentially a breaking change for people who have CS bugs.

pyoptsparse/pyOpt_utils.py
```diff
@@ -529,3 +529,13 @@ def _csc_to_coo(mat: dict) -> dict:
     coo_data = np.array(data)
 
     return {"coo": [coo_rows, coo_cols, coo_data], "shape": mat["shape"]}
+
+
+def _broadcast_to_array(name: str, value, n_vars: int, allow_none: bool = True):
```
nwu63, line 534: I'm not sure if defaulting `allow_none = True` is sensible, I guess the logic is that by default, `broadcast_to_array` should just broadcast and not do type checking. Happy to change this though.

nwu63: Also not sure if this function should go here. Realistically, this module so far has been a linear algebra library for sparse matrices, so we could move that into a new module and keep things that are truly general utils here. Can also do that separately.

pyoptsparse/pyOpt_optimization.py
marcomangano: I was wondering the same thing, could not figure out why this was here in the first place. I can see how this could break things if something is not working on the user side. When we make a new release, we should highlight this!

pyoptsparse/pyOpt_optimization.py
marcomangano: Agreed, looks like a copy-paste leftover. `None` does not make sense here, I agree with using the `allow_none=False` for this broadcast

pyoptsparse/pyOpt_utils.py
nwu63: I believe there is no change in behaviour, since we iterate over the object so it doesn't matter if it's an array or list containing `None`.

pyoptsparse/pyOpt_utils.py
nwu63: 1. Sounds good, I will default to `False` and update everything accordingly.
2. I'll leave that as a potential TODO for later then. I'm okay with things as is.

ArshSaja:
Looks good to me. Can you add a simple test? like the one, you mentioned above?

ArshSaja:
> Looks good to me. Can you add a simple test? like the one, you mentioned above?

actually, NVM. the available tests must go through these changes.

ArshSaja (approved):

marcomangano (approved):

PR #351, opened 2023-08-15
Make nsga2 tests optional + docs update

### LINKED ISSUE #345: NSGA2 tests should be optional
## Description

NSGA2 is now an optional optimizer, since #300. Even though we provide the source, we only install it if SWIG is available. As a result, NSGA2-related tests should be skipped if it is not available.

nwu63:
## Purpose

Closes #345. Also updated docs slightly + added an nsga2 test to the Rosenbrock problem.

## Expected time until merged

1 week

## Type of change

- [x] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/tests/test_nsga2_multi_objective.py
+++ b/tests/test_nsga2_multi_objective.py
@@ -3,7 +3,6 @@
 # Standard Python modules
 import sys
 import unittest
-import warnings
 
 # External modules
 from numpy.testing import assert_allclose
@@ -19,6 +18,7 @@ from testing_utils import OptTest
 class TestNSGA2(OptTest):
     name = "quadratic"
     optName = "NSGA2"
+    histFileName = None
 
     def setUp(self):
         try:
@@ -76,7 +76,4 @@ class TestNSGA2(OptTest):
 
 
 if __name__ == "__main__":
-    unittest.main()
-    unittest.main()
-    unittest.main()
     unittest.main()
--- a/tests/test_nsga2_multi_objective.py
+++ b/tests/test_nsga2_multi_objective.py
@@ -20,6 +20,13 @@ class TestNSGA2(OptTest):
     name = "quadratic"
     optName = "NSGA2"
 
+    def setUp(self):
+        try:
+            # First party modules
+            from pyoptsparse.pyNSGA2 import nsga2  # noqa: F401
+        except ImportError:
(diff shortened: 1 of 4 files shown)


sseraj (approved):

nwu63:
Bump @lamkina

tests/test_nsga2_multi_objective.py
```diff
@@ -43,6 +50,9 @@ def setup_optProb(self, n_obj):
         if n_obj == 2:
             self.optProb.addObj("obj2")
 
+    @unittest.skipIf(
```
marcomangano, line 53: I am sorry to have checked this too late @nwu63 , but I don't really get the point of this step here. Can't we remove this altogether?  We are already skipping all `n_obj == 2` tests by default on all platforms. Is this related to an additional test that `testflo` runs on windows?
Locally I only have the last two tests of this stack:
```
about to run test_nsga2_multi_objective.py:TestNSGA2.test_opt .\test_nsga2_multi_objective.py:TestNSGA2.test_opt  ... SKIP (00:00:0.00, 0 MB)
test_nsga2_multi_objective.py fails on windows with two objectives! Skipping for now.
    about to run test_nsga2_multi_objective.py:TestNSGA2.test_opt_0 .\test_nsga2_multi_objective.py:TestNSGA2.test_opt_0 ... OK (00:00:4.17, 0 MB)
    about to run test_nsga2_multi_objective.py:TestNSGA2.test_opt_1 .\test_nsga2_multi_objective.py:TestNSGA2.test_opt_1  ... SKIP (00:00:0.00, 0 MB)
skip flaky NSGA2 tests
```

nwu63: This skips all tests on Windows, the below skips all multiobjective tests, so the only tests being run are the single-objective tests on linux.

marcomangano: I understand the idea (and that is what I was expecting) but that printout in my comment is from the Windows test log - apparently the single-objective test is running on Windows? Or is it a Testflo fluke?

nwu63: It is possible that the skipIf doesn't play nicely with parameterized. _However_ this is actually what we want---the single objective test should pass on Windows and we should test it, while the multi-objective test appears to be flat out broken. Let me throw together a PR to update this.

Issue #387, opened 2024-02-27
Error while importing pyoptsparse


AbhijithMoni:
### Steps to reproduce issue

`from pyoptsparse.pySLSQP import slsqp`
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "packages/pyoptsparse/pyoptsparse/__init__.py", line 3, in <module>
    from .pyOpt_history import History
  File "packages/pyoptsparse/pyoptsparse/pyOpt_history.py", line 12, in <module>
    from .pyOpt_utils import EPS
  File "packages/pyoptsparse/pyoptsparse/pyOpt_utils.py", line 23, in <module>
    from .types import ArrayType
  File "packages/pyoptsparse/pyoptsparse/types.py", line 9, in <module>
    NumpyType = Union[float, npt.NDArray[np.float64]]
AttributeError: module 'numpy.typing' has no attribute 'NDArray'

### Code versions
v2.10.2 Latest
- Operating System: Linux
- Python: 3.9.5
- Numpy: 1.20.3
- CGNS: 4.3.0 
- PETSc: 3.15.1

marcomangano:
Hi Abhijith,
this looks like a numpy version issue. We don't test numpy 1.20.x anymore (see our supported versions  [here](https://mdolab-mach-aero.readthedocs-hosted.com/en/latest/installInstructions/install3rdPartyPackages.html)), upgrading to numpy 1.21.x should fix this.

AbhijithMoni:
@marcomangano Thanks, Marco. The issue has been resolved.

nwu63:
@marcomangano we should probably update the version specification in `setup.py` though, perhaps following NEP29?

marcomangano:
> @marcomangano we should probably update the version specification in `setup.py` though, perhaps following NEP29?

I think something like NEP29 should be set as a standard over the whole codebase (stay tuned for that), but for now I will open a quick PR to update the minimum numpy version.

PR #388, opened 2024-02-27
Updating minimum dependency requirements

marcomangano:
## Purpose

Issues like #387 show that the current minimum dependency versions are out of date - even considering our own [stencil of supported dependencies](https://mdolab-mach-aero.readthedocs-hosted.com/en/latest/installInstructions/install3rdPartyPackages.html#supported-dependency-versions).
I updated the minimum numpy and scipy versions to our `stable` set of dependencies.

However, I think we should also update the `mdolab-baseclasses` version and, more importantly, the minimum python version. The code itself is not changing but this would create issues for the users who are still relying on older python environments.

@ewu63 suggests to adopt the [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) (or the more recent [SPEC  0](https://scientific-python.org/specs/spec-0000/) policy) to avoid this issue, but I think at that point we should extend this to the whole organization packages.
I want to hear @eirikurj take as well as he has been working on a way to map these dependencies over time.

## Expected time until merged

1 day if we are ok with only updating numpy and scipy

## Type of change

- [x] Breaking change (non-backwards-compatible fix or feature)
- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


sseraj (approved):
Just updating numpy and scipy for now is fine with me

eirikurj:
@marcomangano just to formally respond and for future reference. I am in favor of SPEC 0 for our codebase. We have been discussing this in https://github.com/mdolab/docker/issues/241 and should continue there.

marcomangano:
Bumped the minor version since the minimum version update could potentially be a breaking change?

marcomangano:
@sseraj @lamkina  @eirikurj  can you check this out to see if got the right package versions? We should merge ASAP and make a new release.

sseraj (approved):
This looks good

eirikurj (approved):
Looks good. ~~Just to clarify, do you want to wait for the docker PR?~~ Nervermind, no need to wait.

PR #389, opened 2024-03-17
Better import error

### LINKED ISSUE #187: Improve reporting of 3rd party library import errors
# Description of feature

The idea would be to transmit more information about the error that occurs when an error is encountered importing third party dependencies, in particular `ParOpt` and `IPOPT`.

If this is something the maintainers are interested in I'm happy to flesh this out in a PR.

## Reason for it

At the moment to debug import errors after installation I've had to change:

https://github.com/mdolab/pyoptsparse/blob/b78a64277866c90f995b6a97d8e3cf059e734127/pyoptsparse/pyParOpt/ParOpt.py#L21-L26

to something like this:

```python
try:
        from paropt import ParOpt as _ParOpt
        from mpi4py import MPI
    except ImportError as errid:
        print(errid)
        _ParOpt = None

```

Which lets me realise that I'm missing not finding the import libraries (for example).

# Potential solution

So looking at a more generic solution I'd suggest capturing the error in the module scope and then explicitly re-raising here:

https://github.com/mdolab/pyoptsparse/blob/b78a64277866c90f995b6a97d8e3cf059e734127/pyoptsparse/pyParOpt/ParOpt.py#L44-L46

This would provide a more instructive traceback with the original import error also repeated. An alternative would be to include the original error message into the error raised on lines 44-46.

Thanks for the great package!

ewu63:
## Purpose

Closes #187 

## Expected time until merged

Few days

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing

No new tests, no new change in behaviour other than `ImportError` is now raised instead of generic `Error`.

## Checklist

### DIFF
--- a/pyoptsparse/pyALPSO/pyALPSO.py
+++ b/pyoptsparse/pyALPSO/pyALPSO.py
@@ -10,6 +10,7 @@ import time
 import numpy as np
 
 # Local modules
+from . import alpso
 from ..pyOpt_error import Error
 from ..pyOpt_optimizer import Optimizer
 
@@ -25,9 +26,7 @@ class ALPSO(Optimizer):
     - pll_type -> STR: ALPSO Parallel Implementation (None, SPM- Static, DPM- Dynamic, POA-Parallel Analysis), *Default* = None
     """
 
-    def __init__(self, raiseError=True, options={}):
-        from . import alpso
-
+    def __init__(self, options={}):
         self.alpso = alpso
 
         category = "Global Optimizer"
--- a/pyoptsparse/pyNLPQLP/pyNLPQLP.py
+++ b/pyoptsparse/pyNLPQLP/pyNLPQLP.py
@@ -2,11 +2,6 @@
 pyNLPQLP - A pyOptSparse wrapper for Schittkowski's NLPQLP
 optimization algorithm.
 """
-# Compiled module
-try:
-    from . import nlpqlp  # isort: skip
-except ImportError:
-    nlpqlp = None
 # Standard Python modules
 import datetime
 import os
@@ -18,6 +13,11 @@ import numpy as np
 # Local modules
 from ..pyOpt_error import Error
 from ..pyOpt_optimizer import Optimizer
+from ..pyOpt_utils import try_import_compiled_module_from_path
+
+# import the compiled module
+THIS_DIR = os.path.dirname(os.path.abspath(__file__))
+nlpqlp = try_import_compiled_module_from_path("nlpqlp", THIS_DIR)
 
 
 class NLPQLP(Optimizer):
@@ -30,9 +30,8 @@ class NLPQLP(Optimizer):
         category = "Local Optimizer"
         defOpts = self._getDefaultOptions()
         informs = self._getInforms()
-        if nlpqlp is None:
-            if raiseError:
-                raise Error("There was an error importing the compiled nlpqlp module")
+        if isinstance(nlpqlp, str) and raiseError:
+            raise ImportError(nlpqlp)
 
         super().__init__(name, category, defaultOptions=defOpts, informs=informs, options=options)
         # NLPQLP needs Jacobians in dense format
--- a/pyoptsparse/pyOpt_utils.py
+++ b/pyoptsparse/pyOpt_utils.py
@@ -603,7 +603,6 @@ def try_import_compiled_module_from_path(module_name: str, path: Optional[str] =
         if path is not None:
             warnings.warn(
                 f"{module_name} module could not be imported from {path}.",
-                ImportWarning,
                 stacklevel=2,
             )
         module = str(e)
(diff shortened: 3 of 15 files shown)


pyoptsparse/pyALPSO/pyALPSO.py
```diff
@@ -25,9 +26,7 @@ class ALPSO(Optimizer):
     - pll_type -> STR: ALPSO Parallel Implementation (None, SPM- Static, DPM- Dynamic, POA-Parallel Analysis), *Default* = None
     """
 
-    def __init__(self, raiseError=True, options={}):
-        from . import alpso
-
+    def __init__(self, options={}):
```
VascoSch92, line 29: I will not use an empty dictionary as default value for options here (see [here](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python))

ewu63: This comment has nothing to do with this PR.

pyoptsparse/pyOpt_utils.py
```diff
@@ -570,3 +574,39 @@ def _broadcast_to_array(name: str, value: ArrayType, n_values: int, allow_none:
...
+    if path is not None:
+        path = os.path.abspath(os.path.expandvars(os.path.expanduser(path)))
+        sys.path = [path]
+    try:
+        module = importlib.import_module(module_name)
+    except ImportError as e:
+        if path is not None:
+            warnings.warn(
+                f"{module_name} module could not be imported from {path}.",
+                ImportWarning,
```
marcomangano, line 606: Not related directly to this PR, but is `testflo` raising `ImportWarning` by default? I tested just this function locally and the string above (which can be pretty useful) is not printed out by default - see [here](https://docs.python.org/3/library/warnings.html#warning-categories)

marcomangano (approved):
LGTM, I only have one clarifying question below

pyoptsparse/pyALPSO/pyALPSO.py
VascoSch92: Yes true but you could also solve this problem :-) 

Or open another PR where you solve that. Because it can give bugs :-)

pyoptsparse/pyOpt_utils.py
ewu63: Hmm this is a good point. Maybe I will switch to `UserWarning`? I hate how a lot of warnings are not displayed by default.

marcomangano: Yeah, that's the default but we can set it explicitly.

marcomangano (dismissed):

marcomangano (approved):

pyoptsparse/pyOpt_utils.py
```diff
@@ -570,3 +574,38 @@ def _broadcast_to_array(name: str, value: ArrayType, n_values: int, allow_none:
...
+    except ImportError as e:
+        if path is not None:
+            warnings.warn(
+                f"{module_name} module could not be imported from {path}.",
+                stacklevel=2,
+            )
+        module = str(e)
+    finally:
+        sys.path = orig_path
+    return module
```
lamkina, line 611: Is there a reason why we delay throwing the `ImportError`? If we can't import the module, why not throw the error with the descriptive message here instead of a warning?  

Do we need this "soft fail" type behavior somewhere that I missed?  Otherwise I'm in favor of just throwing the `ImportError` here with the same message and only returning a module if the import is successful.  But again, let me know if I missed something that makes this necessary.

pyoptsparse/pyNLPQLP/pyNLPQLP.py
```diff
@@ -30,9 +30,8 @@ def __init__(self, raiseError=True, options={}):
         category = "Local Optimizer"
         defOpts = self._getDefaultOptions()
         informs = self._getInforms()
-        if nlpqlp is None:
-            if raiseError:
-                raise Error("There was an error importing the compiled nlpqlp module")
+        if isinstance(nlpqlp, str) and raiseError:
```
lamkina, line 33: Related to my other comment, why check for strings and then throw the same error that we could throw in the `try_import_compiled_module_from_path` function?

pyoptsparse/pyOpt_utils.py
ewu63: Yeah it's because this is executed as part of the init process, and we don't necessarily want to raise an error. But if they try to instantiate the class and the compiled library is missing, we throw the error.

pyoptsparse/pyNLPQLP/pyNLPQLP.py
ewu63: This is in the class init, so it's only executed when the class is instantiated.

pyoptsparse/pyOpt_utils.py
lamkina: Okay that makes sense.  I forgot that python tries to import anything that's involved in the init process.  I'll approve this now.

lamkina (approved):

PR #392, opened 2024-04-18
Fix SNOPT printout encoding and skip printing import warnings

### LINKED ISSUE #379: SNOPT printout contains non-UTF8 characters
## Description

This was addressed previously in #51 but is now appearing again. Need to look into the issue.

### Steps to reproduce issue

1. Run an optimization

### Current behavior
The section on Objective, RHS, Ranges and Bounds all display invalid characters

### Expected behavior
The section on Objective, RHS, Ranges and Bounds should display nothing or only dashes.

### Code versions

- SNOPT 7.7.7

ewu63:
## Purpose

Closes #379.
Turns out we just forgot to initialize the `cw` array to `" "`. Also can get rid of `cdummy` now.

Also cleaned up import warnings for compiled libraries -- they are now no longer printed by default at import time for non-default libraries.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


ewu63:
bump @sseraj @lamkina

pyoptsparse/pyOpt_utils.py
```diff
@@ -600,7 +605,7 @@ def try_import_compiled_module_from_path(module_name: str, path: Optional[str] =
     try:
         module = importlib.import_module(module_name)
     except ImportError as e:
-        if path is not None:
+        if raise_warning:
```
marcomangano, line 608: What are the scenarios where we don't want to raise a warning when the import fails? Even if it's from `THIS_DIR` we should probably raise an error

marcomangano (changes requested):
Can you update the PR title/description to mention the new flag? I am also not sure if this is the way to go, see below.
The `cw` update is good stuff!

pyoptsparse/pyOpt_utils.py
ewu63: Right now on the main branch, running e.g. `example/hs015.py` will emit the following warnings for me:
```
.../pyoptsparse/pyoptsparse/pyIPOPT/pyIPOPT.py:28: UserWarning: pyipoptcore module could not be imported from .../repos/pyoptsparse/pyoptsparse/pyIPOPT.
  pyipoptcore = try_import_compiled_module_from_path("pyipoptcore", THIS_DIR)
.../pyoptsparse/pyoptsparse/pyNLPQLP/pyNLPQLP.py:20: UserWarning: nlpqlp module could not be imported from .../pyoptsparse/pyoptsparse/pyNLPQLP.
  nlpqlp = try_import_compiled_module_from_path("nlpqlp", THIS_DIR)
```

Without the changes in this PR, we always print warnings (at import time) if we cannot import these libraries. Given that they are optional and not installed by pyOptSparse, I don't think we will want this for end users. The PR changes this behaviour so that warnings are not emitted by default. Note that errors are raised at instantiation time if import fails and this PR does not change that behaviour.

marcomangano (approved):
LGTM, reverted the version change because we should wait for #394 before the next release

sseraj (approved):
I can confirm that this fixes the non-UTF8 characters in the SNOPT print file. The import changes also look good.

PR #394, opened 2024-04-23
Populate inform dict for all optimizers

### LINKED ISSUE #393: print (sol) of nsga2 is error
I run the examples/nsga2_multi_objectiva.py, it raise when run `print(sol)`.

I found it caused by 
```
line98 ...  inform_val = self.optInform["value"]
line99 ...  inform_text = self.optInform["text"]
```
in `pyOpt_solution.py`.

Becasue in `pyNSGA2.py`:
```
line183... sol_inform = {}
```
The `sol_inform` doesn't have key `value` and `text`.
I temporarily solved it by replacing `sol_inform = {}`  with `sol_inform = {"value":"","text":""}`.

ewu63:
## Purpose

Closes #393.

## Expected time until merged

1 day.
## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


marcomangano:
The flake8 error is fixed in #392, we need to merge that PR before this one and then make a new bugfix release

ewu63:
@marcomangano I would prefer if we do a patch bump asap rather than waiting for #388 (which I expect will take a bit of time?)

As an aside, I do miss my sudo powers...

marcomangano (approved):

marcomangano:
> @marcomangano I would prefer if we do a patch bump asap rather than waiting for #388 (which I expect will take a bit of time?)

#388 is ready to go once this is merged

sseraj (approved):

pyoptsparse/pyALPSO/pyALPSO.py
```diff
@@ -191,9 +192,7 @@ def objconfunc(x):
             self.optProb.comm.bcast(-1, root=0)
 
             # Store Results
-            sol_inform = {}
-            # sol_inform['value'] = inform
-            # sol_inform['text'] = self.informs[inform[0]]
+            sol_inform = {"value": "", "text": ""}
```
swryan, line 195: Was "value" intentionally set to a string value in these optimizers?

As far as I can tell, this was always meant to be an integer return code, and we have logic that expects that to be the case:

```  
File "/home/swryan/dev/OpenMDAO/openmdao/drivers/pyoptsparse_driver.py", line 680, in run
    if exit_status > 2:
       ^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'str' and 'int'
```

Just wondering what the circumstances would be where this value would be a string...

swryan: Hmmm.. I see that we were checking if "value" was in the dict.. and if it is there we expect it to be an int....

So we were handling the case for the optimizers that didn't populate it...  I guess we should consider an empty string value to be an indication that the optimizer doesn't provide this...

ewu63: Yeah I see, I guess we need to agree on a convention for optimizers that don't have any informs. I'm happy to do whatever you guys want, which could be
* `None`
* empty string (current)
* some default inform that just says optimization completed

In the longer term, we should probably make dataclasses and type annotate these better.

ewu63: In retrospect this PR is probably not the best implementation, and was done purely to make the solution object printing work without changing the logic there...

swryan: yeah, it may be that just updating [pyOpt_solution.py](https://github.com/mdolab/pyoptsparse/blob/main/pyoptsparse/pyOpt_solution.py#L98-L99) to only print values for those keys if they exist might a preferable solution, keeping the semantics/typing a bit more rigorous, but it's easy enough for us to handle the empty string (which we will do).

PR #402, opened 2024-05-29
WIP: add unconstrained tests for all optimizers

ewu63:
## Purpose

So a bunch of these tests do not pass...

## Type of change

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

- [ ] I have run `flake8` and `black` to make sure the Python code adheres to PEP-8 and is consistently formatted
- [ ] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable
- [ ] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #403, opened 2024-05-29
flush streams and files before/after optimization

ewu63:
## Purpose

The idea is that, 
1. we flush `stdout` and `stderr` before calling the optimizer, in case prints from the runscript are buffered
2. we flush the output files (which could also be streams) after the end of the optimization, so that stuff printed in the runscript after do not get mixed.

I know we can and should do this for all optimizers, I actually tried to refactor some stuff into "pre-optimization" and "post-optimization" functions shared across all optimizers. But, I stumbled across #402 which means I couldn't actually verify what I was doing with the `dummyConstraint` was not introducing bugs. We can come back to it after that stuff is all cleaned up.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pyoptsparse/pySNOPT/pySNOPT.py
+++ b/pyoptsparse/pySNOPT/pySNOPT.py
@@ -2,10 +2,12 @@
 pySNOPT - A variation of the pySNOPT wrapper specifically designed to
 work with sparse optimization problems.
 """
+
 # Standard Python modules
 import datetime
 import os
 import re
+import sys
 import time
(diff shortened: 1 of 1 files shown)


ewu63:
@sseraj @lamkina

sseraj (approved):

marcomangano (approved):
LGTM, I will take a look at the other PR, let's talk offline

PR #404, opened 2024-05-31
Added restartDict to snstop

sseraj:
## Purpose

I added an option to save `restartDict` after each major iteration. This is helpful when you want to restart optimizations after they crash unexpectedly or run out of time. In the latter case, this avoids the guesswork of setting a conservative time limit. I also added an option that allows `restartDict` to be passed to `snstop_handle` if the user wants to access this after every major iteration.

## Expected time until merged

1-2 weeks

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I added a test that writes `restartDict` using `snstop` and uses the dictionary to restart the optimization. I also tested this with my aerodynamic shape optimization cases.

## Checklist


pyoptsparse/pySNOPT/pySNOPT.py
```diff
@@ -663,12 +663,22 @@ def _snstop(self, ktcond, mjrprtlvl, minimize, n, nncon, nnobj, ns, itn, nmajor,
                 if "funcs" in self.cache.keys():
                     iterDict["funcs"].update(self.cache["funcs"])
 
+        # Create the restart dictionary to be passed to snstop_handle
+        restartDict = {
```
ewu63, line 667: TODO for the future: this should probably be a dataclass object instead of a dict... That way we can avoid some code duplication with above.

ewu63 (commented):
Looks good, like you said this is a breaking change but probably not too bad for downstream users. Another design pattern I had considered previously was for the user to tell pyOptSparse which information to request back from snstop, and those are returned (either as a single dict or just a bunch of args). That would be the most flexible but will require some further implementation.

pyoptsparse/pySNOPT/pySNOPT.py
VascoSch92: @ewu63  I can do it if you open an issue with the specification of what you would like to have ;-)

ewu63: Opened #405

ewu63:
Maybe also do a version bump here (minor?)

pyoptsparse/pySNOPT/pySNOPT.py
sseraj: Thanks for the comments. It would be nice for the function to be more flexible. I'm not sure how returning values would work though because the user does not call `_snstop`. It seems like the dictionaries have to be passed in.

ewu63:
> Thanks for the comments. It would be nice for the function to be more flexible. I'm not sure how returning values would work though because the user does not call `_snstop`. It seems like the dictionaries have to be passed in.

The way I'm thinking is the same as how we implemented `Save major iteration variables`. Basically, pass in a list of what you want, and snstop will be called with those args, matching the order. By default we can always pass `iterDict`.

Another thought - we can also make the "save work arrays at every iter" an option, instead of asking the user to make the snstop function. We can add an option `save work arrays` which by default is False or None, but you can set a path and then the built-in `snstop` function can store these (in some preferred format). This will standardize the restarts.

sseraj:
This is ready for another review. I implemented @ewu63's suggestions so this is more useable and is no longer a breaking change.

tests/test_hs015.py
```diff
@@ -193,6 +195,87 @@ def test_snopt_snstop(self):
         # we should get 70/74
         self.assert_inform_equal(sol, optInform=74)
 
+    @staticmethod
+    def my_snstop_restart(iterDict, restartDict):
+        # Save the restart dictionary
+        writePickle("restart.pickle", restartDict)
```
ewu63, line 201: Can refactor `restart.pickle` to a variable to avoid potential inconsistency

ewu63: Might be easier to move this function inside the test function itself.

tests/test_hs015.py
```diff
@@ -193,6 +195,87 @@ def test_snopt_snstop(self):
...
+
+    def test_snopt_snstop_restart(self):
+        # Run the optimization for 5 major iterations
+        self.optName = "SNOPT"
+        self.setup_optProb()
+        optOptions = {
+            "snSTOP function handle": self.my_snstop_restart,
+            "snSTOP arguments": ["restartDict"],
+        }
+        sol = self.optimize(optOptions=optOptions, storeHistory=True)
```
ewu63, line 217: I would check that this opt exited with 74.

ewu63 (commented):
I think the version bump got swallowed in the merge resolution. Otherwise LGTM, thanks!

tests/test_hs015.py
```diff
@@ -193,6 +195,90 @@ def test_snopt_snstop(self):
...
+            self.optName,
+            options={
+                "Start": "Hot",
+                "Verify level": -1,
+            },
+        )
+        histFile = "work_arrays_save.hst"
+        sol = opt(self.optProb, sens=self.sens, storeHistory=histFile, restartDict=restartDict)
+
+        # Check that the optimization converged in fewer than 5 more major iterations
```
marcomangano, line 274: We are not using `my_snstop_restart` here, and the previous optimization was stopped after just one iteration so this might take more than 5 anyway. Can we remove the comment?

tests/test_hs015.py
```diff
@@ -193,6 +195,90 @@ def test_snopt_snstop(self):
...
+
+        # Check that the optimization exited with 74
+        self.assert_inform_equal(sol, optInform=74)
+
+        # Read the restart dictionary pickle file saved by snstop
+        restartDict = readPickle(pickleFile)
+
+        # Now optimize again but using the restart dictionary
+        self.setup_optProb()
+        opt = OPT(
```
marcomangano, line 228: You are not using `self.optimize()` just to avoid all the extra checks or am I missing some initialization step?

marcomangano (commented):
Just a couple of minor comments, looks great. I will bump the minor version in a minute, but I would like to have #410 merged before this so we can include that in the new release.

tests/test_hs015.py
sseraj: Fixed, thanks

tests/test_hs015.py
sseraj: `self.optimize()` does not have a way of passing in `restartDict`

marcomangano (approved):

ewu63 (approved):

ewu63:
So I forgot that we made it not a breaking change, meaning we should've bumped patch and not minor version... oh well a bit late now, I'd rather not leave any commits on `main` with incorrect version numbers.

marcomangano:
> So I forgot that we made it not a breaking change, meaning we should've bumped patch and not minor version... oh well a bit late now, I'd rather not leave any commits on `main` with incorrect version numbers.

I bumped the minor version because this is technically a new feature, I think we got a bit loose with version numbers tbh. Not a big deal, but we can be a bit more strict in the future

ewu63:
> > So I forgot that we made it not a breaking change, meaning we should've bumped patch and not minor version... oh well a bit late now, I'd rather not leave any commits on `main` with incorrect version numbers.
> 
> I bumped the minor version because this is technically a new feature, I think we got a bit loose with version numbers tbh. Not a big deal, but we can be a bit more strict in the future

In general I have shied away from semver when I was in the lab, and instead stuck largely to [Numpy's versioning scheme](https://numpy.org/doc/stable/dev/depending_on_numpy.html#understanding-numpy-s-versioning-and-api-abi-stability), for the reasons outlined there. There is nothing inherently wrong with bumping the minor version frequently, but it should carry some weight behind it - otherwise we quickly move to v2.52.0 without actual substantial changes to the software.

PR #407, opened 2024-06-21
Fix `_masterFunc2` fail flag caching and add fail flag identification to IPOPT

eytanadler:
## Purpose

Some optimizers don't simultaneously call `_masterFunc` with the objective and constraint evaluation. In these cases, if the primal fails, the fail flag will only return the accurate value when the cache is not used (the case that actually calls the objective/constraint function). When the cache is used, the fail flag is always returned as `0`. A similar behavior occurs with the gradient evaluation. This PR caches the fail flag to fix this edge case.

Secondly, the pyIPOPT wrapper does not do anything if the `_masterFunc` returns a `True` fail flag (`1`). Even if any of the functions fail, IPOPT will just push on as if nothing happened. A CFD-based optimization case I saw with this was that it would get a mesh failure, immediately evaluate the gradients, go straight back to a mesh failure, and so on. The output file was as if nothing had failed and everything was normal. This is clearly not what should happen. I solved it by returning `np.array(np.NaN)` in the callback functions if `fail == 1`.

There are two changes in particular that I'd like feedback on:

1. There's a certain rare logic branch: the gradients are evaluated for a design variable vector at which the primal has not been evaluated. In this case, `_masterFunc2` calls itself to evaluate the primal first. In the current pyOptSparse implementation, if this primal fails, that failure is totally ignored. I modified it so that if the primal fails in this case, the gradient `_masterFunc2` evaluation will return that failure value, even if the gradient evaluation succeeds. See lines 452--458 and 511-517 of the new code. Is this the right thing to do?
2. The only way I could figure out to tell IPOPT that the evaluation failed is to return `np.array(np.NaN)` in whatever callback function is called. I don't see any more elegant way based on their docs. Although the shape isn't guaranteed to match the function's usual array shape, it appears to work based on my test cases. Does this seem like a reasonable way to solve this problem?

P.S. should I update the patch version?

## Expected time until merged

A week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Run the tests I added. Also try running a simple IPOPT optimization case where some of the functions periodically fail. It should backtrack properly (it would just ignore the fail flag previously). For example:

```python
import numpy as np
from pyoptsparse import Optimization, OPT

iters = 0
def objfunc(xdict):
    x = xdict["xvars"]
    funcs = {"obj": x[0]**2 + np.sin(x[0] - 0.5)}

    # Fail every fourth iteration
    global iters
    fail = iters % 4 == 2
    iters += 1

    return funcs, fail

def sensfunc(xdict, funcsDict):
    x = xdict["xvars"]  # Extract array
    funcsSens = {"obj": {"xvars": [2 * x[0] + np.cos(x[0] - 0.5)]}}
    fail = False
    return funcsSens, fail

# Instantiate Optimization Problem
optProb = Optimization("Optimization problem", objfunc)
optProb.addVarGroup("xvars", 1, "c", value=3, scale=1.0)
optProb.addObj("obj")

# Create optimizer
opt = OPT("IPOPT", options={})
sol = opt(optProb, sens=sensfunc)
```

The IPOPT output should look something like this (note the alpha cutback warnings, which do not appear with the current implementation):
```
iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  9.5984721e+00 0.00e+00 5.20e+00   0.0 0.00e+00    -  0.00e+00 0.00e+00   0
   1  4.4065559e+00 0.00e+00 5.30e+00 -11.0 5.20e+00    -  1.00e+00 1.00e+00f  1
Warning: Cutting back alpha due to evaluation error
   2 -1.9724310e-01 0.00e+00 1.59e+00 -11.0 2.62e+00    -  1.00e+00 5.00e-01f  2
   3 -6.2890484e-01 0.00e+00 3.02e-02 -11.0 5.62e-01    -  1.00e+00 1.00e+00f  1
   4 -6.2907135e-01 0.00e+00 1.51e-03 -11.0 1.05e-02    -  1.00e+00 1.00e+00f  1
Warning: Cutting back alpha due to evaluation error
   5 -6.2907166e-01 0.00e+00 7.55e-04 -11.0 5.53e-04    -  1.00e+00 5.00e-01f  2
   6 -6.2907176e-01 0.00e+00 5.10e-08 -11.0 2.76e-04    -  1.00e+00 1.00e+00f  1
   7 -6.2907176e-01 0.00e+00 1.72e-12 -11.0 1.86e-08    -  1.00e+00 1.00e+00f  1
```

## Checklist

### DIFF
--- a/pyoptsparse/pyIPOPT/pyIPOPT.py
+++ b/pyoptsparse/pyIPOPT/pyIPOPT.py
@@ -91,7 +91,7 @@ class IPOPT(Optimizer):
             "sb": [str, "yes"],
             "print_user_options": [str, "yes"],
             "output_file": [str, "IPOPT.out"],
-            "linear_solver": [str, "ma86"],
+            "linear_solver": [str, "mumps"],
         }
         return defOpts
 
--- a/pyoptsparse/pyIPOPT/pyIPOPT.py
+++ b/pyoptsparse/pyIPOPT/pyIPOPT.py
@@ -217,19 +217,25 @@ class IPOPT(Optimizer):
             # Define the 4 call back functions that ipopt needs:
             def eval_f(x, user_data=None):
                 fobj, fail = self._masterFunc(x, ["fobj"])
-                if fail == 2:
+                if fail == 1:
+                    fobj = np.array(np.NaN)
+                elif fail == 2:
                     self.userRequestedTermination = True
                 return fobj
 
             def eval_g(x, user_data=None):
                 fcon, fail = self._masterFunc(x, ["fcon"])
-                if fail == 2:
+                if fail == 1:
+                    fcon = np.array(np.NaN)
+                elif fail == 2:
                     self.userRequestedTermination = True
                 return fcon.copy()
 
             def eval_grad_f(x, user_data=None):
                 gobj, fail = self._masterFunc(x, ["gobj"])
-                if fail == 2:
+                if fail == 1:
+                    gobj = np.array(np.NaN)
+                elif fail == 2:
                     self.userRequestedTermination = True
                 return gobj.copy()
 
@@ -238,7 +244,9 @@ class IPOPT(Optimizer):
                     return copy.deepcopy(matStruct)
                 else:
                     gcon, fail = self._masterFunc(x, ["gcon"])
-                    if fail == 2:
+                    if fail == 1:
+                        gcon = np.array(np.NaN)
+                    elif fail == 2:
                         self.userRequestedTermination = True
                     return gcon.copy()
 
--- a/pyoptsparse/pyIPOPT/pyIPOPT.py
+++ b/pyoptsparse/pyIPOPT/pyIPOPT.py
@@ -91,7 +91,7 @@ class IPOPT(Optimizer):
             "sb": [str, "yes"],
             "print_user_options": [str, "yes"],
             "output_file": [str, "IPOPT.out"],
-            "linear_solver": [str, "mumps"],
+            "linear_solver": [str, "ma86"],
         }
         return defOpts
 
@@ -161,7 +161,7 @@ class IPOPT(Optimizer):
 
         if len(optProb.constraints) == 0:
             # If the user *actually* has an unconstrained problem,
-            # snopt sort of chokes with that....it has to have at
+            # IPOPT sort of chokes with that....it has to have at
             # least one constraint. So we will add one
             # automatically here:
             self.unconstrained = True
--- a/pyoptsparse/pyOpt_optimizer.py
+++ b/pyoptsparse/pyOpt_optimizer.py
@@ -81,7 +81,7 @@ class Optimizer(BaseSolver):
         self.storeSens: bool = True
 
         # Cache storage
-        self.cache: Dict[str, Any] = {"x": None, "fobj": None, "fcon": None, "gobj": None, "gcon": None}
+        self.cache: Dict[str, Any] = {"x": None, "fobj": None, "fcon": None, "gobj": None, "gcon": None, "fail": None}
 
         # A second-level cache for optimizers that require callbacks
         # for each constraint. (eg. PSQP etc)
@@ -388,6 +388,7 @@ class Optimizer(BaseSolver):
 
                 # Update fail flag
                 masterFail = max(masterFail, fail)
+                self.cache["fail"] = masterFail
 
             # fobj is now in cache
             returns.append(self.cache["fobj"])
@@ -437,6 +438,7 @@ class Optimizer(BaseSolver):
 
                 # Update fail flag
                 masterFail = max(masterFail, fail)
+                self.cache["fail"] = masterFail
 
             # fcon is now in cache
             returns.append(self.cache["fcon"])
@@ -447,10 +449,13 @@ class Optimizer(BaseSolver):
                 # The previous evaluated point is different than the point requested for the derivative
                 # OR this is the first call to _masterFunc2 in a hot started optimization
                 # Recursively call the routine with ['fobj', 'fcon']
-                self._masterFunc2(x, ["fobj", "fcon"], writeHist=False)
+                _, _, fail = self._masterFunc2(x, ["fobj", "fcon"], writeHist=False)
                 # We *don't* count that extra call, since that will
                 # screw up the numbering...so we subtract the last call.
                 self.callCounter -= 1
+                # Update fail flag
+                masterFail = max(masterFail, fail)
+                self.cache["fail"] = masterFail
             # Now, the point has been evaluated correctly so we
             # determine if we have to run the sens calc:
 
@@ -491,6 +496,7 @@ class Optimizer(BaseSolver):
 
                 # Update fail flag
                 masterFail = max(masterFail, fail)
+                self.cache["fail"] = masterFail
 
             # gobj is now in the cache
             returns.append(self.cache["gobj"])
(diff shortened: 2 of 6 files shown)


eytanadler:
Does the Windows action failure have something to do with my changes? It doesn't seem like it, but I'm not very familiar with it

pyoptsparse/pyOpt_optimizer.py
```diff
@@ -544,13 +553,15 @@ def _masterFunc2(self, x, evaluate, writeHist=True):
...
+                self.cache["fail"] = masterFail
 
             # gcon is now in the cache
             returns.append(self.cache["gcon"])
             if self.storeSens:
                 hist["funcsSens"] = self.cache["funcsSens"]
 
-        # Put the fail flag in the history:
+        # Update the fail flag with any cached failure and put the fail flag in the history
+        masterFail = max(self.cache["fail"], masterFail)
```
ewu63, line 564: Can we do a single `self.cache["fail"] = masterFail` call here? Instead of having the same line in each of the `if` blocks

tests/test_optimizer.py
```diff
@@ -0,0 +1,224 @@
...
+        # Finalize
+        self.optProb.printSparsity()
+        # create and store optimizer
+        self.opt = OPT("slsqp", options={"IFILE": "optProb_SLSQP.out"})
+        self.opt(self.optProb, sens="FD")
+
+        # Call the masterFunc with some infinite DVs so it resets iters
+        self.opt._masterFunc(np.full(np.sum(nDV), np.inf), ["fobj"])
+
+    def test_masterFunc_fobj_fail(self):
```
ewu63, line 119: Unit tests! Excellent. Do these all fail without the fix in this PR?

ewu63 (commented):
Good stuff. I had known that the fail flag is not passed to IPOPT correctly (in fact I think SNOPT is the only one that supports it). From my cursory read of the docs passing `NaN` seems reasonable (and is in fact how many other optimizers prefer to handle failed evals.
As to the implementation, it looks good, I just left a minor comment if we can deal with the cache once instead of running the same line in multiple blocks.

As for the Windows failure, I suspect it's related to the recent release of numpy 2.0 (which broke our meson build in multiple ways, I haven't had time to investigate). See [here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=959345&view=logs&jobId=171a126d-c574-5c8c-1269-ff3b989e923d&j=171a126d-c574-5c8c-1269-ff3b989e923d&t=623c2ba8-c4f1-582d-6c78-1c7699e2e0a2) for a similar log. I would suggest pinning numpy<2  in the env file first.

pyoptsparse/pyIPOPT/pyIPOPT.py
```diff
@@ -217,19 +217,25 @@ def __call__(
             # Define the 4 call back functions that ipopt needs:
             def eval_f(x, user_data=None):
                 fobj, fail = self._masterFunc(x, ["fobj"])
-                if fail == 2:
+                if fail == 1:
+                    fobj = np.array(np.NaN)
+                elif fail == 2:
```
VascoSch92, line 222: can we have the parameter fail different from 1 and 2? Should we catch this case and raise an error? Because if this happens then it can raise a bug difficult to detect :-)

pyoptsparse/pyIPOPT/pyIPOPT.py
```diff
@@ -217,19 +217,25 @@ def __call__(
             # Define the 4 call back functions that ipopt needs:
             def eval_f(x, user_data=None):
                 fobj, fail = self._masterFunc(x, ["fobj"])
-                if fail == 2:
+                if fail == 1:
```
VascoSch92, line 220: small note: you are not using the parameter user_data in this method :-)

pyoptsparse/pyOpt_optimizer.py
VascoSch92: and do you need to define the dictionary self.cache already with all the keys? can you just not define an empty dicionary?

tests/test_optimizer.py
eytanadler: Yep :) I added a few more for good measure
```
----------------------------------------------------------------------
Ran 10 tests in 0.032s

FAILED (failures=6)
```

pyoptsparse/pyOpt_optimizer.py
eytanadler: Unfortunately, a single cache set here won't work. In that case, any time it fails, `_masterFunc2` will always return failures since it allows only monotonically increasing fail values (taking the max of the cached and current values every time). I tried a couple other things to avoid duplicate code, but I think that caching the fail value has to happen at the same time the actual user functions are called or else it won't take on the right value.

eytanadler:
> Good stuff. I had known that the fail flag is not passed to IPOPT correctly (in fact I think SNOPT is the only one that supports it). From my cursory read of the docs passing `NaN` seems reasonable (and is in fact how many other optimizers prefer to handle failed evals. As to the implementation, it looks good, I just left a minor comment if we can deal with the cache once instead of running the same line in multiple blocks.
> 
> As for the Windows failure, I suspect it's related to the recent release of numpy 2.0 (which broke our meson build in multiple ways, I haven't had time to investigate). See [here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=959345&view=logs&jobId=171a126d-c574-5c8c-1269-ff3b989e923d&j=171a126d-c574-5c8c-1269-ff3b989e923d&t=623c2ba8-c4f1-582d-6c78-1c7699e2e0a2) for a similar log. I would suggest pinning numpy<2 in the env file first.

Thanks for the speedy review! I pinned the version of NumPy in the setup.py (I assume this is what you meant by env file?) to `<2`, but the Windows action still seems to fail. I also bumped pyOptSparse's patch version because I think this is a substantial enough change that it'd be worth versioning it. @ewu63, let me know if you think otherwise.

pyoptsparse/pyOpt_optimizer.py
ewu63: Yeah I see, that makes sense. Some of the duplicated code can probably be pulled out into another function or something, but not needed right now.

ewu63:
> Thanks for the speedy review! I pinned the version of NumPy in the setup.py (I assume this is what you meant by env file?) to `<2`, but the Windows action still seems to fail. I also bumped pyOptSparse's patch version because I think this is a substantial enough change that it'd be worth versioning it. @ewu63, let me know if you think otherwise.

Windows builds do not use setuptools but instead uses conda. The env file used by GHA is [here](https://github.com/mdolab/pyoptsparse/blob/main/.github/environment.yml). Also fine with the patch version bump.

eytanadler:
> > Thanks for the speedy review! I pinned the version of NumPy in the setup.py (I assume this is what you meant by env file?) to `<2`, but the Windows action still seems to fail. I also bumped pyOptSparse's patch version because I think this is a substantial enough change that it'd be worth versioning it. @ewu63, let me know if you think otherwise.
> 
> Windows builds do not use setuptools but instead uses conda. The env file used by GHA is [here](https://github.com/mdolab/pyoptsparse/blob/main/.github/environment.yml). Also fine with the patch version bump.

Yay it worked!

ewu63 (approved):
Good work!

marcomangano (approved):
Looking great! Thanks for adding the tests, you went beyond what I had in mind. Very elegant setup.

We should double check failure handling with other optimizers and maybe add a test, but that should be a separate PR. The changes in `pyIPOPT` make sense.

eytanadler:
> Looking great! Thanks for adding the tests, you went beyond what I had in mind. Very elegant setup.
> 
> We should double check failure handling with other optimizers and maybe add a test, but that should be a separate PR. The changes in `pyIPOPT` make sense.

Thanks! I agree failure handling should be checked, but I didn’t have an immediate good idea for how to test that. I’m sure there’s a way.

Issue #353, opened 2023-08-24
Update ParOpt Interface


A-CGray:
# Description of feature

ParOpt has seen some recent updates that have changed it's API and added support for general sparse constraints that pyOptSparse could take advantage of. See [here](https://smdogroup.github.io/paropt/notebooks/sparse.html)

# Potential solution

We need to update the ParOpt interface to return sparse and dense constraints, and their jacobians, separately.
As a first attempt, we could simply treat linear constraints as sparse and nonlinear constraints as dense, I think this would already result in a large performance improvement. After that we could look into something like determining which constraints are sparse using the density of each row in the full constraint Jacobian.

PR #409, opened 2024-06-26
Update ParOpt interface to support sparse constraints

A-CGray:
## Purpose

ParOpt now supports sparse constraints which should help some of the performance issues I've run into with it for problems with a lot of constraints.

To use this functionality, the ParOpt problem you create needs to define new methods for computing the sparse constraints separately and another for computing jac-vec products with the sparse constraint jacobian.

As a first step, I figured the best approach would be to treat the linear constraints as sparse and the nonlinear constraints as dense. 

I'm opening this as a draft now as I'm not sure the best way to separate the evaluation of the linear and nonlinear constraints, hoping to get @ewu63 's opinion. In particular, I think ParOpt will call the jac-vec product function quite a lot so it might be important to make that quite efficient, ideally we would just have a single sparse mat for all the linear constraints and then do a single mat-vec product.

Closes https://github.com/mdolab/pyoptsparse/issues/353

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pyoptsparse/pyParOpt/ParOpt.py
+++ b/pyoptsparse/pyParOpt/ParOpt.py
@@ -212,6 +212,39 @@ class ParOpt(Optimizer):
                         A[i][:] = -gcon[i][:]
                     return fail
 
+                def evalSparseCon(self, x, out):
+                    """Evaluate the sparse constraints.
+
+                    Give the design variable vector x, compute the sparse constraints.
+
+                    Parameters
+                    ----------
+                    x : array
+                        Design variable vector
+                    out : array
+                        Sparse constraint values
+                    """
+                    pass
+
+                def addSparseJacobian(self, alpha, x, px, out):
+                    """Compute the Jacobian-vector product of the sparse constraints.
+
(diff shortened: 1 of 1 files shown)


A-CGray:
@gjkennedy just tagging to make you aware of this

A-CGray:
@gjkennedy I see there are two different ways to work with sparse constraints in ParOpt:

1. Implement `evalSparseObjCon` and `evalSparseObjConGradient` so that the sparse constraints and jacobian are returned along with the dense constraints
2. Implement separate `evalSparseCon` and `addSparseJacobian` methods just for evaluating the sparse constraints.

Is either one of these approaches preferable from ParOpt's point of view? Also, in option 1 I see that `evalSparseObjConGradient` directly populates the data array of the ParOpt sparse jacobian, how/when is the sparsity pattern supposed to be given to ParOpt?

A-CGray:
Closing as this is superceded by #414

Issue #296, opened 2022-04-21
Linear Constraints ignore the constant part of the equation.


Kenneth-T-Moore:
## Description

If I provide a constraint defined by `y = m*x + b`, and set linear=True for the ConGroup, it is treated as if it were `y - m*x`. 

### Steps to reproduce issue

I've put together a simple [test here](https://gist.github.com/Kenneth-T-Moore/9160a9df5ba7f5953ab037effaa042db)

The problem has 2 constraints.  One of them has a very high constant term -> `y = 2x + 100`. This term insures there is no feasible solution to the problem. When this constraint is defined with linear=False, SNOPT finds the problem infeasible, which is expected. When linear=True, SNOPT finds a solution (exit 0, info 1), even though the problem is unsolvable. This is because it ignores the constant term in the equation. You can verify by changing the value of LINEAR near the top of the test file.

### Current behavior

I think what is happening is that pyoptsparse's knowledge of this constraint comes entirely from the jacobian we give it in `addConGroup`.  We never give it an initial value, nor does it ever ask for the current value when it calls the objfun. From the jacobian, it can only evaluate the equation as `y = m*x`, which is incorrect.

### Expected behavior

This is obviously a simplified, nonsensical problem, but I expect that it should fail to find a solution if the linear constraint was working as expected. This can be accomplished as follows:

When LINEAR=True, require an initial value for the constraint. That way, pyoptsparse can back-calculate the correct y-intercept for the constraint. The initial value should correspond to the initial design variable values that are given to pyoptsparse.  

### Code versions

- Operating System:
- Python:
- OpenMPI:
- CGNS:
- PETSc:
- Compiler:
- This repository:

nwu63:
I believe in pyOptSparse, the `b` part of the constraint is supplied as part of the bound information. For example, if you have a constraint `c = 2*x + 100` and you want to constrain `c` between -10 and 10, you will have to call `addCon` with `lower=-110` and `upper=90`. Then it should work.

kanekosh:
I'm considering adding the y-intercept `b` as an optional input to `addConGroup`. If `b` is provided, we impose `LB < Ax + b < UB`, and otherwise `LB < Ax < UB` by default.
[...]

PR #410, opened 2024-06-26
Added linear constraint check and updated documentation

kanekosh:
## Purpose

pyOptSparse computes the linear constraints exclusively based on `jac`, `lower`, and `upper` entries of `addConGroup`, and it ignores the linear constraint value returned by the user-defined obj/con function. This is confusing and can result in unexpected behavior (see related #296).

This PR adds a linear constraint checkerto make sure that a user-defined obj/con function does **not** return linear constraint values. If it does, pyOptSparse will raise an error. I also updated the documentation accordingly.

I fixed a few existing tests that have been computing linear constraint values in the obj/con function (which we should not). I also added a test to check if pyOptSparse raises an error as expected.

This change will break some existing runscripts that return linear constraint values in the user-defined function. OpenMDAO's pyOptSparseDriver will be fine. Not sure about MACH runscripts (haven't checked).

## Type of change

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/tests/test_snopt_bugfix.py
+++ b/tests/test_snopt_bugfix.py
@@ -21,35 +21,6 @@ def objfunc(xdict):
     funcs = {}
 
     funcs["obj"] = (x - 3.0) ** 2 + x * y + (y + 4.0) ** 2 - 3.0
-    conval = -x + y
-    funcs["con"] = conval
-
-    fail = False
-    return funcs, fail
-
-
-def objfunc_no_con(xdict):
-    """Evaluates the equation f(x,y) = (x-3)^2 + xy + (y+4)^2 - 3"""
-    x = xdict["x"]
-    y = xdict["y"]
-    funcs = {}
-
-    funcs["obj"] = (x - 3.0) ** 2 + x * y + (y + 4.0) ** 2 - 3.0
-
-    fail = False
-    return funcs, fail
-
-
-def objfunc_2con(xdict):
-    """Evaluates the equation f(x,y) = (x-3)^2 + xy + (y+4)^2 - 3"""
-    x = xdict["x"]
-    y = xdict["y"]
-    funcs = {}
-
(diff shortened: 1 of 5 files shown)


tests/test_snopt_bugfix.py
```diff
@@ -21,35 +21,6 @@ def objfunc(xdict):
...
 
     funcs["obj"] = (x - 3.0) ** 2 + x * y + (y + 4.0) ** 2 - 3.0
-    conval = -x + y
-    funcs["con"] = conval
-
-    fail = False
-    return funcs, fail
-
-
-def objfunc_no_con(xdict):
```
marcomangano, line 31: Nice catch, I assume the new check had these tests failing right away, right?

marcomangano (approved):
No additional comments on the check we came up with, thanks for fixing the tests!!

tests/test_snopt_bugfix.py
kanekosh: Yes you're right

marcomangano:
@lamkina @ewu63 @sseraj can you check and merge this soon so we can go ahead and make a new release?

lamkina:
Do we have a unit test that confirms the standard way of adding linear constraints is working as expected? If not, can that be added to this PR? If it exists and I missed it, then point me to the test and I'll approve this PR.

This would just be as simple as adding a few linear constraints to a valid opt prob and then making sure the number of linear constraints and the Jacobian are correct.

kanekosh:
I don't know if there's a test that checks the number of linear constraints, but we already have a few tests that include linear constraints and assert equal on the optimal solution, which validates that linear constraints are properly imposed.

lamkina:
That's good enough for me. Thanks for the double check!

lamkina (approved):

Issue #303, opened 2022-06-04
Consider updating SLSQP to a modern and maintained implementation


nwu63:
# Description of feature
Currently, the version of SLSQP provided is quite old, and suffers from several bugs that have been fixed elsewhere. See #301 for some discussion. Since SLSQP remains a rather popular optimizer, to maintain long term viability, I think it would be best to switch to using a version that is better maintained. This would also avoid any duplication in maintenance efforts.

# Potential solution
As far as I'm aware, there are three versions out there:
- Scipy: well maintained and widely available, plus we already depend on scipy so there will be no additional dependencies. However seems to lack things such as fetching the optimal Lagrange multipliers that exist in pyOptSparse (though what we have might be broken, I don't really remember)
- slsqp: much more modern than the old F77 code, seems to be very well maintained. Lacks Python interface
- NLopt: given that it's built into an entire optimization framework, we will not consider this option further

This thread will serve as a place to discuss future plans regarding SLSQP.

andrewellis55:
@ewu63 Question for you on this - does pyOptSparse add any measurable advantage over just using SciPy minimize for SLSQP? If the end result is using a dense Jacobean anyway. I understand the usefulness of the common interface between various optimizers, but as an OpenMDAO user where the ScipyOptimizeDriver is available, in your opinion is there any advantage to using pyOptSparse for SLSQP?

ewu63:
> @ewu63 Question for you on this - does pyOptSparse add any measurable advantage over just using SciPy minimize for SLSQP? If the end result is using a dense Jacobean anyway. I understand the usefulness of the common interface between various optimizers, but as an OpenMDAO user where the ScipyOptimizeDriver is available, in your opinion is there any advantage to using pyOptSparse for SLSQP?

It's not clear to us - we did some benchmarking years ago and did not find conclusively whether our version or SciPy's version is superior. @marcomangano can add more on this probably. Like you said, key features are generally not supported since SLSQP is a very old and little-maintained optimizer that has been pretty static, so I don't expect performance to differ too much between versions, and the bounds issue is the only notable divergent aspect that I am aware of.
If you are interested in advanced features such as sparsity or failure handling, IPOPT or SNOPT is probably your best bet.

marcomangano:
[...]

PR #413, opened 2024-09-14
Strongly enforce SLSQP Bounds

### LINKED ISSUE #301: SLSQP does not strictly follow problem bounds
I am running into a lot of issues because the SLSQP algorithm does not strictly follow bounds. In multiple cases this has resulted in the optimization terminating with values outside of bounds even when the initial values were feasible and in bounds. From what I've been able to find elsewhere this may be a known issue. If so is there a hack around this? And if not this seems to me like a pretty significant issue.

andrewellis55:
## Purpose
One of the key differences between the pyOptSparse SLSQP implementation and the SciPy SLSQP implementation is strong bound enforcement. This was added the the SciPy implementation in these two PRs

https://github.com/scipy/scipy/pull/4502
https://github.com/scipy/scipy/pull/13009

These changes are fairly small and easy enough to port over. I've proposed this with this PR. This I suppose should fix #301 and relate to #303.

## Expected time until merged
This is not overly pressing, but my team uses OpenMDAO with the SciPy SLSQP rather than the pyOptSparse driver due to bound enforcement issues, so this will allow us to shift over to pyOptSparse once merged.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
I ran the hs015.py example and brought the initial guess outside to the bounds.

## Checklist

### DIFF
--- a/pyoptsparse/pySLSQP/pySLSQP.py
+++ b/pyoptsparse/pySLSQP/pySLSQP.py
@@ -7,7 +7,7 @@ work with sparse optimization problems.
 import datetime
 import os
 import time
-import warnings
+from ..pyOpt_error import pyOptSparseWarning
 
 # External modules
 import numpy as np
@@ -168,8 +168,8 @@ class SLSQP(Optimizer):
             # =================================================================
             def slfunc(m, me, la, n, f, g, x):
                 if (x < blx).any() or (x > bux).any():
-                    warnings.warn("Values in x were outside bounds during a "
-                                  "minimize step, clipping to bounds", RuntimeWarning)
+                    pyOptSparseWarning("Values in x were outside bounds during"
+                                       " a minimize step, clipping to bounds")YY
                 fobj, fcon, fail = self._masterFunc(np.clip(x, blx, bux), ["fobj", "fcon"])
                 f = fobj
                 g[0:m] = -fcon
--- a/pyoptsparse/pySLSQP/pySLSQP.py
+++ b/pyoptsparse/pySLSQP/pySLSQP.py
@@ -169,7 +169,7 @@ class SLSQP(Optimizer):
             def slfunc(m, me, la, n, f, g, x):
                 if (x < blx).any() or (x > bux).any():
                     pyOptSparseWarning("Values in x were outside bounds during"
-                                       " a minimize step, clipping to bounds")YY
+                                       " a minimize step, clipping to bounds")
                 fobj, fcon, fail = self._masterFunc(np.clip(x, blx, bux), ["fobj", "fcon"])
                 f = fobj
                 g[0:m] = -fcon
--- a/pyoptsparse/pySLSQP/pySLSQP.py
+++ b/pyoptsparse/pySLSQP/pySLSQP.py
@@ -7,6 +7,7 @@ work with sparse optimization problems.
 import datetime
 import os
 import time
+import warnings
 
 # External modules
 import numpy as np
@@ -166,6 +167,9 @@ class SLSQP(Optimizer):
             # SLSQP - Objective/Constraint Values Function
             # =================================================================
             def slfunc(m, me, la, n, f, g, x):
+                if (x < blx).any() or (x > bux).any():
+                    warnings.warn("Values in x were outside bounds during a "
+                                  "minimize step, clipping to bounds", RuntimeWarning)
                 fobj, fcon, fail = self._masterFunc(np.clip(x, blx, bux), ["fobj", "fcon"])
                 f = fobj
                 g[0:m] = -fcon
--- a/pyoptsparse/pySLSQP/pySLSQP.py
+++ b/pyoptsparse/pySLSQP/pySLSQP.py
@@ -220,6 +220,10 @@ class SLSQP(Optimizer):
             # fmt: on
             optTime = time.time() - t0
 
+            # Clip final result to user bounds (this occurs during the optimization as well
+            # so this just makes the output consistent with what the optimizer sees)
+            xs = np.clip(xs, blx, bux)
+
             # some entries of W include the lagrange multipliers
             # for each constraint, there are two entries (lower, upper).
             # if only one is active, look for the nonzero. If both are active, take the first one
--- a/pyoptsparse/pySLSQP/pySLSQP.py
+++ b/pyoptsparse/pySLSQP/pySLSQP.py
@@ -166,7 +166,7 @@ class SLSQP(Optimizer):
             # SLSQP - Objective/Constraint Values Function
             # =================================================================
             def slfunc(m, me, la, n, f, g, x):
-                fobj, fcon, fail = self._masterFunc(x, ["fobj", "fcon"])
+                fobj, fcon, fail = self._masterFunc(np.clip(x, blx, bux), ["fobj", "fcon"])
                 f = fobj
                 g[0:m] = -fcon
                 slsqp.pyflush(self.getOption("IOUT"))
@@ -176,7 +176,7 @@ class SLSQP(Optimizer):
             # SLSQP - Objective/Constraint Gradients Function
             # =================================================================
             def slgrad(m, me, la, n, f, g, df, dg, x):
-                gobj, gcon, fail = self._masterFunc(x, ["gobj", "gcon"])
+                gobj, gcon, fail = self._masterFunc(np.clip(x, blx, bux), ["gobj", "gcon"])
                 df[0:n] = gobj.copy()
                 dg[0:m, 0:n] = -gcon.copy()
                 slsqp.pyflush(self.getOption("IOUT"))
--- a/pyoptsparse/pySLSQP/pySLSQP.py
+++ b/pyoptsparse/pySLSQP/pySLSQP.py
@@ -7,12 +7,12 @@ work with sparse optimization problems.
 import datetime
 import os
 import time
-from ..pyOpt_error import pyOptSparseWarning
 
 # External modules
 import numpy as np
 
 # Local modules
+from ..pyOpt_error import pyOptSparseWarning
 from ..pyOpt_optimizer import Optimizer
 from ..pyOpt_utils import try_import_compiled_module_from_path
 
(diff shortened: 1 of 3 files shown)


ewu63:
Hey, thanks for the PR and for fixing a long-standing issue with our SLSQP implementation. Could you clarify whether SLSQP will generate intermediate points which violate bounds, or if the final solution can violate the bounds also?

The high-level code looks fine but I haven't looked too closely yet. In the mean time, can I ask that you add a test to show the intended fix (i.e. a test that fails without the code changes here)? You can probably base it off of an existing test in `test_h015.py` or `test_hs071.py`

andrewellis55:
Thanks for taking a look @ewu63, I've added a simple test that just raises a `ValueError` if the optimizer ever steps outside of the bounds. With the case I added, the `main` branch hits this `ValueError`, the new branch does not.

Also good point on ensure the clipping occurs to the final solution as well. I've added a few lines to ensure this and checked for it in the test I added.

whophil:
Any opinions about making this behavior configurable with an option flag?

marcomangano:
Thanks for the bugfix @andrewellis55! I will also take a closer look in the next few days, the implementation makes sense at a first pass. I believe the tests are currently failing because the code is missing a `self.optName = "SLSQP"` line in the `test_slsqp_strong_bound_enforcement` function.

@whophil I don't believe that there are scenarios where we want the optimizer to violate the design variables bounds, do you have an example in mind?
I think this bugfix is worth a new patch release anyway, we can highlight the fix / change in behavior in the release notes.

akleb:
I am concerned about the behavior of clipping when you are trying to take steps with Hessian information on multi-dimensional problems.

If we have a slanted quadratic like the example unconstrained case in the [MDO book](https://mdobook.github.io/) (D.1.1), clipped Newton steps will not get the minimum within the design variable bounds if you have Hessian information. If you have the Hessian, you will always be taking steps exactly to the global optimum regardless of where you are in the design space. This means you will end up with the state pushed up against the bound aligned with the global optimum (purple in the attached image). If instead, you did not use Hessian information, the gradient would push you down to the correct bounded optimum where the gradient is perpendicular to the design space boundary (red in the attached figure).

If you are building an approximate Hessian I am not sure what would happen, but I am guessing you would have some sort of in between behavior. However, for a simple case like this, I think you would build the exact (linearized?) Hessian within one or two steps, so if you did not start exactly on the bounded optimum, you would not get the correct solution.

To solve this you have to either switch to a lower dimensional problem on bounds or include the design variable bounds in the Lagrangian. In my experience the better more robust solution is to include the design bounds in the Lagrangian, I think there are a lot of potential pitfalls in switching to a lower dimensional problem for problems with multimodality resulting in cycling behavior. I can describe my thoughts on this with more pictures if desired.

It could be that the design variables bounds are already included in the QP and some other weird behavior is causing the design variables to exceed their bounds. If this is the case, perhaps clipping is an okay solution. Regardless, I think that providing a multidimensional test case that shows we avoid this behavior would be beneficial. Also having this issue is much better than a solution that violates design variable bounds.

IMG_0101

andrewellis55:
@akleb You raise some good points. First of all, as other have mentioned, the optimizer generally should never step outside these bounds, meaning that as long as all goes well, the clipping should never occur and thus the hessian approximation should be unaffected. I've ran the example problem you've provided from the MDO book with the code as follows

```python
from pyoptsparse import Optimization, SLSQP
import numpy as np


def slanted_quadratic(xdict):
    x1 = xdict['x1']
    x2 = xdict['x2']
    beta = 3/2  # beta = 3/2 as shown in the image
    f = x1**2 + x2**2 - beta * x1 * x2  # Function (D.1)
    funcs = {}
    funcs['f'] = f
    return funcs, False

def slanted_quadratic_grad(xdict, funcs):
    x1 = xdict['x1']
    x2 = xdict['x2']
    beta = 3/2
    grad_f = np.array([2*x1 - beta * x2, 2*x2 - beta * x1])
    grad = {}
    grad['f'] = {'x1': grad_f[0], 'x2': grad_f[1]}
    return grad, False

# Set up the optimization problem
optProb = Optimization('Slanted Quadratic Minimization', slanted_quadratic)
optProb.addVar('x1', 'c', lower=-10.0, upper=10.0, value=-9)
optProb.addVar('x2', 'c', lower=-10.0, upper=-5.0, value=9)
optProb.addObj('f')

# Define the optimizer
optimizer = SLSQP()

# Solve the optimization problem
sol = optimizer(optProb, sens=slanted_quadratic_grad)

# Print the results
print(sol)
```

I've added a print statement before the clipping at the python level in `slfunc` to plot the path of the variables

```python
            # =================================================================
            # SLSQP - Objective/Constraint Values Function
            # =================================================================
            def slfunc(m, me, la, n, f, g, x):
                print(x)
                fobj, fcon, fail = self._masterFunc(np.clip(x, blx, bux), ["fobj", "fcon"])
                f = fobj
                g[0:m] = -fcon
                slsqp.pyflush(self.getOption("IOUT"))
                return f, g
```

Here is the trajectory
```
1: [-9. -5.]
2: [ 1.5 -8.5]
3: [-5.37931034 -6.20689655]
4: [-3.85207101 -5.        ]
5: [-3.75571076 -5.        ]
6: [-3.75 -5.  ]
```
Figure_1

Although it isn't clear in the image that this is the minima along the line, this can be easily verified

```python
>>> def slanted_quadratic(x1, x2, beta=1.5):
...     return x1**2 + x2**2 - beta * x1 * x2
>>> slanted_quadratic(-5, -3.74, beta=1.5)
10.9376
>>> slanted_quadratic(-5, -3.75, beta=1.5)
10.9375
>>> slanted_quadratic(-5, -3.76, beta=1.5)
10.9376
```

As can be seen , in this example SLSQP actually never exceeds the bounds and behaves normally. There have been edge cases that I (as well as other users #301 ) have seen where this is not the case for a particular iteration. An example of this that I've personally run is a optimizer of a tube with design variables `inner_diameter` and `thickness` and a `safety_margin` constraint that is essentially a `1/x` function of force. The optimizer has stepped the thickness to be negative (exceeding the lower bounds) and caused the optimization to fail as the gradients of the margin function point in the opposite direction on the left side of the asymtote.

andrewellis55:
@kanekosh Both I and other user (#301) have observed SLSQP exceeding bounds. This was also observed in the SciPy implementation leading to this fix https://github.com/scipy/scipy/pull/4502. My code is a pure copy paste of the fortran changes and a slight adaptation of the python ones that were implemented in SciPy.

As for why it's happening, I frankly havn't read through the details of the code enough to say, but if I were to hazard a guess based off my general knowledge of SQP algorithms, small bound violations could come from numerical issues and larger bound violations could come if the bounds are treated as general non-linear constraints and/or are not included in the active set.

The fix I've added is the one that SciPy has been running for the past ~9 years and so I think it's generally a fairly safe one.

kanekosh:
@andrewellis55 Would you mind adding lines of code to raise a warning when the clipping occurs, like scipy does [here](https://github.com/scipy/scipy/blob/97445cc6057a45ce871ddb4087157330646539be/scipy/optimize/_optimize.py#L308)?

There might be a scenario where this clipping would confuse the optimization algorithm, so raising a warning would be beneficial just in case. (The benefit of enforcing the bounds is much greater than the potential convergence issue by clipping, so I think this is a good fix and I appreciate it)

whophil:
> @whophil I don't believe that there are scenarios where we want the optimizer to violate the design variables bounds, do you have an example in mind?

@marcomangano my question was primarily about the ability to enable "falling back" to the previous implementation of the SLSQP algorithm without having to install an older version of`pyoptsparse`.

andrewellis55:
> @andrewellis55 Would you mind adding lines of code to raise a warning when the clipping occurs, like scipy does [here](https://github.com/scipy/scipy/blob/97445cc6057a45ce871ddb4087157330646539be/scipy/optimize/_optimize.py#L308)?

@kanekosh Done!

andrewellis55:
> > @whophil I don't believe that there are scenarios where we want the optimizer to violate the design variables bounds, do you have an example in mind?
> 
> @marcomangano my question was about enabling "fallback" to the previous implementation of the SLSQP algorithm without having to install an older version of`pyoptsparse`.

While that's probably simple enough on the python side, it would require a bit of modification on Fortran wrapper side too which would made this PR a bit more than a copy/paste job

pyoptsparse/pySLSQP/pySLSQP.py
```diff
@@ -166,7 +167,10 @@ def __call__(
             # SLSQP - Objective/Constraint Values Function
             # =================================================================
             def slfunc(m, me, la, n, f, g, x):
-                fobj, fcon, fail = self._masterFunc(x, ["fobj", "fcon"])
+                if (x < blx).any() or (x > bux).any():
+                    warnings.warn("Values in x were outside bounds during a "
```
kanekosh, line 171: This line [broke the build](https://dev.azure.com/mdolab/Public/_build/results?buildId=8201&view=logs&j=2d879094-0a86-54e9-c117-83eecb3ade7b&t=28b2e3e5-a4a9-54fc-8c3d-6c70517ff4c0&l=93), would you mind fixing it?
Also, could you use the [pyOptSparseWarning class](https://github.com/mdolab/pyoptsparse/blob/main/pyoptsparse/pyOpt_error.py) instead to clarify that this warning comes from pyOptSparse?

andrewellis55: Done! But just a note, there may be some cases where those might get a little verbose as it could get raised quite a few times compares to the python warning class that I believe shouldn't re-trigger

image

kanekosh (approved):

marcomangano (approved):
Thanks again for the fix, I think this is a welcome improvement that addresses a long-standing bug. I don't see the need for further tweaks of the source code - if we had to include all the possible improvements we would probably end up with a substantially different code, or a different optimizer altogether. Moving forward, we should consider the discussion in #303 and switch to an external maintained version.

PR #414, opened 2024-10-02
Add dense and sparse wrappers for ParOpt from ParOpt directly

gjkennedy:
## Purpose

This PR adds the ParOpt sparse and dense constraint wrappers for pyOptSparse from ParOpt directly.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


A-CGray:
@marcomangano @ewu63 the actual pyoptsparse wrapper that this will end up using is [here](https://github.com/smdogroup/paropt/blob/472750988fcc5d3c019735c75c17443f0c0d4c08/paropt/paropt_pyoptsparse.py)

A-CGray:
The tests are failing because our docker images are pinned to an older version of ParOpt. But if we update the ParOpt version in the docker images before merging this PR then the image builds will fail and so the tests here will still fail. So I guess we need to test this PR locally, merge it even with failing azure tests, and then update the docker images. Opinions @ewu63 @eirikurj ?

A-CGray:
@gjkennedy I've run the pyoptsparse tests locally, the dense version of your wrapper seems to work fine, but when using the sparse version the constraints seem to be ignored. Try running `test_hs015.py`

marcomangano:
Specific test errors aside, I have mixed feelings about moving the actual pyOptSparse wrapper to the parOpt repo.

On the one hand, outsourcing it could make things a bit easier for us in terms of maintenance, especially if the SMDO group plans to update the python API on a regular basis.
On the other hand, if we make breaking changes to our codebase, we would not be able to address it directly and have to wait for a new version release of ParOpt for our tests to pass. Since we run tests on our side, I am leaning towards moving the updated wrapper back here.

This would also avoid the kind of "circular import" happening now, where pyoptsparse imports paropt but the imported wrapper is also importing pyoptsparse components -  but this is a minor note.
Curious to hear @ewu63 and @eirikurj take on this.

ewu63:
1. I have no idea how this circular dependency would be resolved, and in any case I don't think pyoptsparse-specific code should live in ParOpt. Instead, a generic Python interface should be provided. I think the correct approach would be the following, consistent with our other wrappers
    1. Keep the wrapper in this repo
    2. Test with specific tagged version, and request that changes to API be versioned appropriately
2. Once the wrapper is confirmed to be working, we can merge this PR and then update docker. The procedure you described is fine @A-CGray

A-CGray:
@gjkennedy we can go with this approach, but there are currently some issues with the wrapper in the ParOpt. I opened https://github.com/smdogroup/paropt/pull/51 to fix them. With those fixes I can get the paropt tests in this repo to pass. Once that PR is merged and you make a new release of ParOpt, we can figure out how to get things working in our docker images.

@ewu63 , the MPI environment variable tests below are now failing because this PR removed [this code](https://github.com/mdolab/pyoptsparse/blob/f66e9fbd6712dbb941794698cf87fc053c8a7986/pyoptsparse/pyParOpt/ParOpt.py#L13-L29), I had a look at adding it back but I can't understand the logic behind the code, could you take a look?

https://github.com/mdolab/pyoptsparse/blob/f66e9fbd6712dbb941794698cf87fc053c8a7986/tests/test_require_mpi_env_var.py#L45-L73

pyoptsparse/pyParOpt/ParOpt.py
```diff
@@ -1,263 +1,24 @@
...
+from pyoptsparse.pyOpt_optimizer import Optimizer
+
+try:
+    # External modules
+    from paropt.paropt_pyoptsparse import ParOptSparse as ParOpt
+except ImportError:
+
+    class ParOpt(Optimizer):
+        def __init__(self, raiseError=True, options={}):
+            name = "ParOpt"
```
VascoSch92, line 12:
```suggestion
        def __init__(self, raiseError=True, options=None):
            options = options if options else {}
            name = "ParOpt"
```

Why? -> https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil

A-CGray: Because this is the standard defined by the optimizer base class 

https://github.com/mdolab/pyoptsparse/blob/f66e9fbd6712dbb941794698cf87fc053c8a7986/pyoptsparse/pyOpt_optimizer.py#L34-L36

Feel free to open a PR making this fix for all optimizers

A-CGray:
Marking this as a draft until https://github.com/smdogroup/paropt/pull/51 is fixed and a new ParOpt release is available

A-CGray:
There is one more PR on ParOpt (https://github.com/smdogroup/paropt/pull/55) that needs to be merged to fix some small remaining issues and then the v2.1.5 release of ParOpt updated to the latest commit. Even once that is done, the tests here will continue to fail because the docker images use an old version of ParOpt. Previously we said we'd test this PR locally, merge it, then update the ParOpt version in the docker images.

I added a new set of tests that tests the three different ParOpt algorithms with both the dense and sparse versions of the wrapper on a constrained and unconstrained version of the Rosenbrock function. I also updated the tests related to the MPI environment variable.

marcomangano:
[This test](https://github.com/gjkennedy/pyoptsparse/blob/d36d430a85e3f48b641528ceacb3e33486e9c9b9/tests/test_require_mpi_env_var.py#L45-L76) is failing because the docker images are not updated to the latest `paropt` version, but I believe we could remove this test altogether at this point?

A-CGray:
> Hey @A-CGray , I'm not entirely sure of the behavior we're checking for here, relative to OpenMDAO. If ParOpt tests are being moved out of ParOpt I think I'm ok with this test being removed.

Sorry, I wasn't very clear, specifically we're asking whether OpenMDAO still relies on the behaviour added by Herb in https://github.com/mdolab/pyoptsparse/pull/118 where pyOptSparse can be required to use or not use MPI using the `PYOPTSPARSE_REQUIRE_MPI` environment variable.

kanekosh (approved):

PR #416, opened 2024-11-13
IPOPT now returns Lagrange multpliers using the same sign convention as SNOPT.

### LINKED ISSUE #415: Lagrange multiplier printout in summary output is broken.
## Description

When printing out a summary string of a solution, the Lagrange multipliers are always `9E100` even when they are provided by the optimizer.

The cause of the bug is as follows.  `Solution` is a subclass of `Optimization`.  When printing the solution using `str(sol)`, it first calls `super.__str__()`.

Within `Optimization.__str__`, the multipliers are only included on the following condition:

```
if not isinstance(self, Optimization) and self.lambdaStar is not None:
```

However, for both `Solution` and `Optimization` the `isinstance` call will be `True`.

### Steps to reproduce issue

Running the hs071.py example with IPOPT (or SNOPT)

### Current behavior

```
python hs071.py --opt=IPOPT
```

will print the constraints as

```
   Constraints (i - inequality, e - equality)
      Index  Name Type          Lower           Value           Upper    Status  Lagrange Multiplier (N/A)
          0  con1    i   2.500000E+01    2.500000E+01    1.000000E+20         l    9.00000E+100
          1  con2    e   4.000000E+01    4.000000E+01    4.000000E+01              9.00000E+100
```

### Expected behavior

When working correctly, the results should be something like:

```
   Constraints (i - inequality, e - equality)
      Index  Name Type          Lower           Value           Upper    Status  Lagrange Multiplier
          0  con1    i   2.500000E+01    2.500000E+01    1.000000E+20         l     5.52294E-01
          1  con2    e   4.000000E+01    4.000000E+01    4.000000E+01              -1.61469E-01
```

### Code versions
- pyOptSparse:  2.12.0
(This is purely an issue with the python source)

robfalck:
## Purpose

Added lagrange multipliers to IPOPT's output. Note that in the call to _createSolution we negate the multiplier values so that they are consistent with those returned by SNOPT. My understanding is that this is a matter of convention (due to the internal representations of the problems in the two optimizers).

The lagrange multipliers in the string representation of the solution are now correct.

Resolves #415 

## Expected time until merged

This change is not urgent.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I've added a test to hs071 that verifies that the lagrange multipliers returned by SNOPT and IPOPT are correct.

## Checklist

### DIFF
--- a/pyoptsparse/pyIPOPT/pyIPOPT.py
+++ b/pyoptsparse/pyIPOPT/pyIPOPT.py
@@ -296,7 +296,7 @@ class IPOPT(Optimizer):
             sol_inform["text"] = self.informs[status]
 
             # Create the optimization solution
-            sol = self._createSolution(optTime, sol_inform, obj, x, multipliers=-constraint_multipliers)
+            sol = self._createSolution(optTime, sol_inform, obj, x, multipliers=constraint_multipliers)
 
             # Indicate solution finished
             self.optProb.comm.bcast(-1, root=0)
--- a/pyoptsparse/pyIPOPT/pyIPOPT.py
+++ b/pyoptsparse/pyIPOPT/pyIPOPT.py
@@ -296,7 +296,7 @@ class IPOPT(Optimizer):
             sol_inform["text"] = self.informs[status]
 
             # Create the optimization solution
-            sol = self._createSolution(optTime, sol_inform, obj, x)
+            sol = self._createSolution(optTime, sol_inform, obj, x, multipliers=-constraint_multipliers)
 
             # Indicate solution finished
             self.optProb.comm.bcast(-1, root=0)
--- a/pyoptsparse/pyOpt_optimization.py
+++ b/pyoptsparse/pyOpt_optimization.py
@@ -1577,7 +1577,7 @@ class Optimization:
         con_opt = self._mapContoOpt(con)
         return self.processContoDict(con_opt, scaled=False, natural=True)
 
-    def summary_str(self, minimal_print=False):
+    def summary_str(self, minimal_print=False, print_multipliers=False):
         """
         Print Structured Optimization Problem
 
@@ -1588,6 +1588,8 @@ class Optimization:
             variables and constraints with a non-empty status
             (for example a violated bound).
             This defaults to False, which will print all results.
+        print_multipliers : bool
+            If True, print the Lagrange multipliers associated with the constraints.
         """
         TOL = 1.0e-6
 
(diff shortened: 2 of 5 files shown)


marcomangano:
Thanks for opening the PR! I also got curious and worked out a logic fix but you were quicker than me - see this [comparison](https://github.com/mdolab/pyoptsparse/compare/main...multipliers-print-fix). I don't have strong opinions between that approach and just add a flag as you did, I actually like yours a bit more because it is more explicit. @lamkina do you have any suggestions here?

It looks like the tests are failing because Paropt also returns multipliers, so you have to add it to the new test. I believe you are using a different setting for black, we allow 120-characters-long lines, hence the formatting fails.

As you see from my branch, I also figured that the IPOPT convention is flipped and changed the sign of the multipliers, so I am 100% onboard with what you did.

robfalck:
I'm not sure where the regression in coverage is coming in.  As far as I can tell it's hitting all of the lines I added.  It might be due to black reformatting a few of the things.

robfalck:
You're right that my initial use of black was a little too agressive (no 120 char limit specification).  I dialed that back so it only affected the files that I needed to change.

I'm still failing codecov and would appreciate guidance on what you'd like me to do.

marcomangano:
Sorry I just noticed your issues with Codecov - you actually don't need to do anything there! Since the PR comes from a fork and not from `mdolab`, the tests are ran on the publicly available Docker images that do not include SNOPT and NLPQLP, hence the reduction in coverage. Not passing the coverage test does not block the merge.

marcomangano (dismissed):
I verified that the missing tests pass locally, this looks good to me, thanks!

ewu63:
This is a bit of a philosophical question, but do we have references that show IPOPT uses a different sign convention from SNOPT? I vaguely recall SNOPT itself using the opposite sign of normal convention, but it's been a while.

If we are modifying values that are coming out of the optimizers, this needs to be clearly documented. I'm not entirely against it, but this can easily confuse downstream users and create a mess. If we want to ensure that all Lagrange multipliers are consistent across optimizers, then I think what we need then is the following:
1. A pyOptSparse definition of Lagrange multipliers with the sign that we will use
2. A note in the documentation page for each optimizer where we flip the sign that is output from the optimizers
3. A clear note/warning explaining this in the docstring of `lambdaStar` field
4. A minor version bump

robfalck:
I think it's fair to move slowly with defining lambdaStar that is consistent across optimizers. I will update the PR to return the native IPOPT lambdaStar, since we can handle that case downstream and we don't necessarily want to confuse other users.
The main intent of this PR was to enable the return of lagrange multipliers from IPOPT and to have them print correctly (for all optimizers) in the results summary table. There's no need to hold that up for a more philosophical question.

marcomangano (approved):

ewu63 (approved):

PR #417, opened 2024-11-22
parallel snstop

awccopp:
## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Any script that runs in parallel check pass snSTOP handle and check if all mpi ranks call the function and have the same input arguments.

## Checklist

### DIFF
--- a/.github/test_real.sh
+++ b/.github/test_real.sh
@@ -11,4 +11,4 @@ cd tests
 # we have to copy over the coveragerc file to make sure it's in the
 # same directory where codecov is run
 cp ../.coveragerc .
-testflo --pre_announce --disallow_deprecations -v --coverage --coverpkg pyoptsparse $EXTRA_FLAGS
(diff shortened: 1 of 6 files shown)


marcomangano:
@awccopp about the tests (that won't run here because the PR originates from a fork and not `mdolab`), we should find a way to either modify [this function](https://github.com/awccopp/pyoptsparse/blob/b028be45b11d7f7eeff69e727510cba8ed9e43e5/tests/test_hs015.py#L181-L185) or create a new one to do operations in parallel. The tricky issue is that we are not calling MPI or setup comms anywhere in the test file right now. Maybe we could have a new test file only for parallel tests? @eirikurj do you have any suggestions?

eirikurj:
@marcomangano Yeah, without having thought too much about this, for now I think a second parallel test file would be the best. We can think about restructuring tests later to be more general.
@awccopp can you please update the PR description with the relevant details, such as what the issue is and what this PR intends to solve.

marcomangano:
Merging this to an `mdolab` branch so we can run SNOPT tests from there.

Issue #256, opened 2021-05-13
Add more test coverage


nwu63:
# Description
## Unit tests
- [ ] test that the Solution object contains the right info
	- [x] `objectives`
	- [ ] `constraints`
	- [ ] `variables`

## Integration tests
- [ ] MPI tests
- [ ] other optimization problems?

PR #420, opened 2024-12-17
parallel snstop

awccopp:
## Purpose

The purpose of this PR is to parallelize the SNSTOP user callback function.
See #417 for commit history.
A `--timeout` option has been added to the `testflo` arguments to avoid tests to hang.

## Expected time until merged

1 month

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Tests were added to ensure other processors are calling the snstop function.

## Checklist


marcomangano (dismissed):

marcomangano:
@ewu63 you talked with @awccopp about this PR, but might still want to take a look?
@eirikurj does this look good to you?

lamkina (approved):
This looks good to me. I would still wait for @ewu63 or @eirikurj to have a look before merging if possible.

marcomangano (approved):
It looks like it is finally working. Is 120s a reasonable timeout timeout for tests?

marcomangano:
I see your point @ewu63 . Would that be as simple as adding an additional bool option (say `parallelSnStop`) and then only run [these lines](https://github.com/mdolab/pyoptsparse/blob/feee7a6e90312056b6d1f582d617d192a5f7c522/pyoptsparse/pySNOPT/pySNOPT.py#L746-L748) if `'parallelSnStop is True'`?

The rest of the edits in the `_waitLoop()` function are just to make the `mode` variable more explicit (before it was just checking for the `-1` flag to break the wait), and add the parallel option just within pySNOPT.

ewu63:
I thought about this PR some more and I have some more reservations
* we currently don't have any MPI-based tests and those probably need to be added first to make sure we are not causing any regressions. #256 lists some of the tests I had in mind
* the MPI imports need to be adjusted. There is an implicit contract with OpenMDAO and the wider community that mpi4py is not imported unless instructed -- this is why we have the special pyOpt_MPI module. So the internal code needs to be changed here.

marcomangano:
> I can try to address some of this soon but I would prefer holding off on this PR for a bit longer. Trying to be diligent to not break people's stuff downstream.

Agreed, there is no rush. I can add that flag we mentioned soon, so that the default behavior is unchanged. Down to brainstorm additional tests.

PR #425, opened 2025-03-12
Switch to cyipopt

### LINKED ISSUE #374: Linking pyoptsparse to cyipopt
### Discussed in https://github.com/mdolab/pyoptsparse/discussions/373

<div type='discussions-op-text'>

<sup>Originally posted by **pollinico** December 26, 2023</sup>
Hi, I just installed PyOptSparse with SNOPT, and I am very excited about how well everything works! Thank you, guys!
Just a thought: Have you considered linking the installation of PyOptSparse to [cyipopt](https://github.com/mechmotum/cyipopt)? 
In this way, it would be easier to include IPOPT in the PyOptSparse environment. 
I don't know what this would entail; it goes beyond my current capabilities, but it would make everything much easier, I think. 
Please let me know what you think. 
Thank you for your attention, Nicolo.</div>

ewu63:
## Purpose
Closes #374, and hopefully addresses some of the issues we've been having. I also added the long-awaited numpy2 test on CI (with windows).

Some to dos:
- [x] decide whether we want to specify cyipopt as a dependency
- [x] check options, setups, etc. to cyipopt and compare against previous code
- [x] update documentation

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


ewu63:
@kanekosh are you still using IPOPT? It would be nice if you can try out the changes here and see if they give similar results as before

kanekosh:
Thanks @ewu63! I tested cyipopt (both one from conda-force and pip-installed one with my pre-build IPOPT) and it works almost identical to my previous IPOPT installation (I sometimes see very minor difference as a results of different IPOPT and MUMPS versions).

One minor fix: I happened to have a bad optimization problem where a constraint function call fails and IPOPT does "Cutting back alpha due to evaluation error" and goes on (and magically converges after that). When the evaluation error occurs, we used to return `np.array(np.NaN)` regardless of the number of constraints and that made IPOPT do alpha cut back. Now with cyipopt, returning `np.array(np.NaN)`  causes an unexpected error in cyipopt [because of these lines](https://github.com/mechmotum/cyipopt/blob/097fae9b6f950c68f28a9de37e36750057728f32/cyipopt/cython/ipopt_wrapper.pyx#L1021), and cyipopt sends user termination signal instead of passing NaNs to IPOPT. 
Instead, we need to pass the correct shape of NaNs under the evaluation failure. This array of NaNs goes into IPOPT and triggers alpha cut back as we intend. I just pushed a change to do so.

I believe it's the same for gradient and jacobian callbacks, so I made changes to return array of NaNs. However I haven't tested these so not 100% sure...

pyoptsparse/meson.build
```diff
@@ -66,7 +66,7 @@ inc_f2py = include_directories(incdir_f2py)
 #)
 
 subdir('pySNOPT')
-subdir('pyIPOPT')
+# subdir('pyIPOPT')
```
kanekosh, line 69: Should we remove this line?

pyoptsparse/pyALPSO/meson.build
```diff
@@ -1,13 +1,13 @@
-python_sources = [
-    '__init__.py',
-    'alpso.py',
-    'alpso_ext.py',
-    'pyALPSO.py',
-    'LICENSE'
-]
+# python_sources = [
```
kanekosh, line 1: Are these comment outs intended?

pyoptsparse/pyIPOPT/meson.build
```diff
@@ -1,60 +1,10 @@
...
-                    'src/pyipoptcoremodule.c',
-                    include_directories: [inc_np, 'src', ipopt_inc],
-                    dependencies : [py3_dep, ipopt_dep],
-                    subdir: 'pyoptsparse/pyIPOPT',
-                    link_language: 'c',
-                    install : false)
-endif
-
-#python_sources = [
+# python_sources = [
```
kanekosh, line 1: Should we delete this meson.build file?

ewu63:
I will cleanup the meson files (either here or in #431). Yeah since we are not going to use meson to install the python files I'll just remove most of the lines.

Do you think you can add a test for the condition you described? I'm surprised we don't have one on eval failure but I suppose that part isn't very well tested.

kanekosh:
Sounds good! Yes I'll can add a test.

pyoptsparse/pyALPSO/meson.build
ewu63: Yeah it doesn't really matter, these files will be gone as part of #431 and this file was never used anyways due to `subdir('pyALPSO')` being commented out. I can revert to keep the git history a bit cleaner

tests/test_hs015.py
```diff
@@ -278,22 +278,83 @@ def test_snopt_work_arrays_save(self):
...
+                funcs["con"] = [np.nan, np.nan]
+                fail = True
+            return funcs, fail
+
+        self.optName = optName
+        self.setup_optProb()
+        # swap obj to simulate eval failure
+        self.optProb.objFun = objFun_eval_failure
+        sol = self.optimize(optOptions={}, storeHistory=True)
+        # sheck solution and informs
```
ewu63, line 354: nit: "sheck"

tests/test_hs015.py
```diff
@@ -278,22 +278,83 @@ def test_snopt_work_arrays_save(self):
...
+        self.assert_solution_allclose(sol, 1e-5)
+        self.assert_inform_equal(sol)
```
ewu63, line 356: Can add another check that, in the history file the 3rd iteration was nan

ewu63:
> Sounds good! Just added a few tests. The new test (`test_func_eval_failure`) currently only tests SNOPT and IPOPT (two most robust optimizers) as other optimizers can fail depending on the environment. SLSQP and PSQP passed the test on my machine but failed on some images. NLPQLP and ParOpt passed on all images so I could add them back, although I'm not sure if we should do so.

This is fine, I only know SNOPT and IPOPT support failed evaluations so I'm not surprised some of the others failed. Left a few small comments but otherwise this looks good to go.

tests/test_hs015.py
kanekosh: I tried doing this but realized that the history file omits failed function evaluations (which was also documented in the hist file page)

marcomangano (approved):
Great work!! I haven't tested the code myself but I went through the changes and they look straightforward. The docs are clear and @kanekosh 's updated tests look good - nice job with the comments.

I wonder if this PR warrants a minor release?

kanekosh (approved):

ewu63:
I personally would prefer not bumping the version here, and instead wait until #431 to bump the minor version together. If we think #431 will take a while then I'm okay with making a minor release here. CC @whophil

marcomangano:
> I personally would prefer not bumping the version here, and instead wait until #431 to bump the minor version together. If we think #431 will take a while then I'm okay with making a minor release here. CC @whophil

That is fine with me!

PR #428, opened 2025-03-29
Improve windows build 

whophil:
## Purpose

This PR improves the apparently flaky Windows build by making it better match the conda-forge build.

Flang is pinned to 5, which is the same as the current conda-forge global pin. The previous builds were using "latest" Flang, which resolved to 19. I'm not sure why the Flang 19 builds were flaky.

Note that conda-forge is currently migrating to Flang 19. As more conda-forge projects migrate to Flang 19, we can consider updating the CI here to Flang 19, leaning on the experience of the increasing number of conda-forge packages using Flang 19.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -10,6 +10,7 @@ dependencies:
   - pip
   - setuptools
   - build
+  - flang 5*
 # runtime
   - packaging
   - mdolab-baseclasses >=1.3.1
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -11,6 +11,7 @@ dependencies:
   - setuptools
   - build
   - flang 5*
+  - libpgmath
 # runtime
   - packaging
   - mdolab-baseclasses >=1.3.1
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -25,11 +25,6 @@ jobs:
           channel-priority: strict
           activate-environment: pyos-build
           environment-file: .github/environment.yml
-      - name: Install libpgmath
-        shell: bash -l {0}
-        run: |
-          conda activate pyos-build
-          conda install libpgmath
       - name: Build and install pyoptsparse
         shell: cmd /C CALL {0}
         run: |
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -36,10 +36,10 @@ jobs:
           call conda activate pyos-build
           set IPOPT_DIR=%CONDA_PREFIX%\Library
 
-          # We want to use the clang linker `ld-lld.exe`
-          # The conda-forge clang packages set the LD environment variable and LDFLAGS for that linker.
-          # But meson doesn't respect the LD environment variable - so set `<compiler entry>_LD`
-          # https://mesonbuild.com/howtox.html#set-linker
+          REM We want to use the clang linker `ld-lld.exe`
+          REM The conda-forge clang packages set the LD environment variable and LDFLAGS for that linker.
+          REM But meson doesn't respect the LD environment variable - so set `<compiler entry>_LD`
+          REM https://mesonbuild.com/howtox.html#set-linker
           set CC_LD=%LD%
           set CXX_LD=%LD%
           set FC_LD=%LD%
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -34,7 +34,6 @@ jobs:
         shell: cmd /C CALL {0}
         run: |
           call conda activate pyos-build
-          set IPOPT_DIR=%CONDA_PREFIX%\Library
 
           :: debug - list conda environment
           call conda list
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -31,6 +31,7 @@ def run_meson_build():
     sysargs = meson_call.split(" ")
     sysargs = [arg for arg in sysargs if arg != ""]
     p1 = subprocess.run(sysargs, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    print(p1.stdout.decode())
```
whophil, line 34: For debugging builds, it is incredibly useful to see the meson configure output even when it succeeds, as it lists the compilers and linkers that were found.

I'm not married to this exactly implementation, but I think it should be _possible_ to easily enable this output without having to patch setup.py.

.github/workflows/windows-build.yml
```diff
@@ -25,26 +25,24 @@ jobs:
...
-          set IPOPT_DIR=%CONDA_PREFIX%\Library
 
-          # We want to use the clang linker `ld-lld.exe`
-          # The conda-forge clang packages set the LD environment variable and LDFLAGS for that linker.
-          # But meson doesn't respect the LD environment variable - so set `<compiler entry>_LD`
-          # https://mesonbuild.com/howtox.html#set-linker
-          set CC_LD=%LD%
-          set CXX_LD=%LD%
-          set FC_LD=%LD%
+          :: debug - list conda environment
```
whophil, line 33: This debug info is very useful. I see no cost to leaving these lines in, as it only affects the windows build logs on CI.

whophil:
A few comments:
1. Not really sure why the coverage is down
2. I am fine with any tweaks the maintainers may want to make to `setup.py` as long as the meson configure/build logs can, perhaps optionally, be redirected to stdout upon success.

setup.py
ewu63: Yeah this is fine with me and is something I have been meaning to do., We do the same for the build step (but currently only if it fails).

.github/environment.yml
```diff
@@ -10,6 +10,8 @@ dependencies:
   - pip
   - setuptools
   - build
+  - flang 5*
```
ewu63, line 13: What is `5*`? Is this equivalent to `flang=5`?

.github/workflows/windows-build.yml
```diff
@@ -25,26 +25,24 @@ jobs:
...
+          call conda list
+
+          :: debug - list environment vars
+          set
+
+          :: set fortran compiler, flang 5 activation doesn't seem to do it
+          set FC=flang.exe
+
+          set MESON_ARGS=-Dipopt_dir=%CONDA_PREFIX%\Library\
```
ewu63, line 42: Any difference between setting `MESON_ARGS` and the previous approach of setting `IPOPT_DIR`? The meson setup should be able to understand either but was just curious why you changed it

setup.py
ewu63: Actually, is the `print(p2.stdout)` duplicate with line 52? Can remove the latter if we want to always print build logs.

ewu63:
Changes look good, just left a few minor comments. I do think in the medium term we should probably move to mesonpy as that project seems mature, at the time we decided to wait because it seems it would get merged back to meson but that no longer appears to be the plan.

.github/workflows/windows-build.yml
whophil: No difference, I just wanted to mimic the conda-forge build as closely as possible.

setup.py
whophil: Fixed

whophil:
> I do think in the medium term we should probably move to mesonpy as that project seems mature

Agree, I think with meson-py a large amout of boilerplate can be outsourced.

marcomangano (approved):
Thanks for fixing this @whophil ! I am onboard with the more verbose printouts. Not familiar with meson-py @ewu63 but I am interested in anything that can make the build process easier to maintain and tweak.

setup.py
```diff
@@ -31,24 +31,22 @@ def run_meson_build():
...
     setup_log = os.path.join(staging_dir, "setup.log")
     with open(setup_log, "wb") as f:
         f.write(p1.stdout)
     if p1.returncode != 0:
-        with open(setup_log, "r") as f:
-            print(f.read())
         raise OSError(sysargs, f"The meson setup command failed! Check the log at {setup_log} for more information.")
```
ewu63, line 39: Could improve the error message slightly by mentioning that it is also printed above. Bonus points with adding a header/footer to `print(p1.stdout)` to clearly delineate which part is from the log. Can do the same with `p2` below.

ewu63 (approved):

setup.py
whophil: Fair points. Given the imminent switch to meson-py (I can work on it), I would also be OK if one of the maintainers wanted to merge as-is.

ewu63:
Alright, merging and then hopefully the other PRs can go through too.

ewu63:
FYI @whophil the coverage went down because this PR was initiated on your branch, and therefore CI skips testing some proprietary optimizers (such as SNOPT).

PR #431, opened 2025-04-02
[Draft] Build with `meson-python` backend

whophil:
## Purpose

Build with [meson-python](https://mesonbuild.com/meson-python/) backend.

This simplifies the build system, allowing for removal of nearly all code in `setup.py`, which is now replaced with `pyproject.toml`. In addition, the `meson.build` files are cleaned up where possible, and updated to use the latest helpful patterns from scipy's meson files.

TODO:
- [x] add dynamic version
- [ ] no IPOPT in Linux builds?

## Type of change

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [ ] Maintenance update
- [x] Other (please describe)

## Testing

Existing tests.

## Checklist

- [ ] I have run `flake8` and `black` to make sure the Python code adheres to PEP-8 and is consistently formatted
- [ ] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable
- [ ] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #433, opened 2025-04-09
Build with mesonpy

ewu63:
## Purpose
Continuation of #431 but from the mdolab branch to test all optimizers.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


ewu63:
Not sure how to do that exactly, the docs aren't very clear. If we can't figure out a way I don't think it's a deal breaker, think it's also been the case since basically the beginning, so I propose that we merge and deal with it in the future.
I did open #434 though and I'd like that merged first, so we can test all the optional installs with mesonpy and make sure that all looks good.

whophil:
@ewu63 this change excludes the `source` and subdir `README` files 0dd27fbd423b1143ccb328a7d3876c09131033d5

ewu63:
> @ewu63 this change excludes the `source` and subdir `README` files [0dd27fb](https://github.com/mdolab/pyoptsparse/commit/0dd27fbd423b1143ccb328a7d3876c09131033d5)

Yeah that looks good, feel free to update this PR

marcomangano:
FYI, the real issue is dug into an error log dropdown printout:
```
  Sanity testing Fortran compiler: flang.exe
  Is cross compiler: False.
  Sanity check compiler command line: flang.exe sanitycheckf.f -o sanitycheckf.exe
  Sanity check compile stdout:
  LINK : fatal error LNK1104: cannot open file 'msvcrt.lib'
  
  -----
  Sanity check compile stderr:
  flang.exe: error: linker command failed with exit code 1104 (use -v to see invocation)
```

I am not familiar with compilers on Windows, but looking at some [discussions like this one](https://stackoverflow.com/questions/22426367/fatal-error-lnk1104-cannot-open-file-msvcrt-lib) (or [this one](https://stackoverflow.com/questions/927574/compiling-c-program-causes-fatal-error-lnk1104) - does the Windows build use Visual Studio in the backend?)we might need to update a compiler option or link to the missing library

kanekosh (dismissed):

marcomangano:
I am sorry I was unclear, the build works just fine, the error above occurs whenever you try and import `pyoptsparse`. I am really confused by the fact that the `_pyoptsparse_editable_loader.py` script generated by meson includes some kind of rebuild on the fly.

ewu63:
I see, it seems that the behaviour here is a little different from `setuptools`: in an editable install, mesonpy will actually check if the compiled code has been modified, and recompiles as necessarily. This is nice in principle because now the compiled library is also considered "editable", instead of requiring manual deletion of build dir + recompiling. However, it does require the following
- running `pip install -e .` with the `--no-build-isolation` flag, since this secondary compilation step must happen in your dev env -- for some reason mesonpy does not create temporarily build envs for subsequent builds
- building without isolation means that your dev env must have the build deps installed (which are specified in `pyproject.toml`). This was the nice thing about setuptools, where it would install build deps for you in a separate build env. This is now done only initially, but in subsequent builds it will have trouble finding the build tools again since that original env has been thrown away.
- This rebuild nicety is unfortunately not disable-able (see [here](https://github.com/mesonbuild/meson-python/issues/730)).

So, I'm not sure what we want to do. I think the dev experience is slightly degraded but not making `os.subprocess` calls in `setup.py` seems like a major improvement. I also buy the argument (from meson devs) that editable installs are inherently not isolated builds, so expecting them to be isolated is pointless. We also already require `gcc` etc. for these installs, so some extra build requirements are not the end of the world if we document these steps properly. What do you all think?

marcomangano:
I need to look into the discussion a bit more in detail, but I think a practical solution could be enabling something like `pip install -e [dev] .` and have the build-time dependencies installed in the same environment "on demand". This would avoid future version headaches with our docker stencil and retain the current editing capabilities. As long as we document this, we should be fine.

Since we only (aside minor exceptions) work with python, it feels weird to recompile (even if it is just through `pip`) as we do code dev, but I agree that we should just find a work around and move on with ditching setuptools as soon as possible.

marcomangano:
EDIT: the `[dev]` approach might also be a problem because we need `numpy >= 2.0` at build time

ewu63:
Circling back on this, what do we think? Is this ready to be merged? @marcomangano @whophil @kanekosh @eirikurj 
I would also probably bump the minor version here.

marcomangano:
Regardless, I think this PR warrants a minor version bump instead. Given the conversation in #459 and the doc update in #466 (https://github.com/mdolab/pyoptsparse/blob/f7b9c68fbb1790a6d6bec897709188165545030d/doc/install.rst?plain=1#L32)  we should probably bump the minimum python version in `pyproject.toml` and `environment.yml`

marcomangano (approved):
Tested both the standard and in-place installation locally (without SNOPT) and everything is working.

ewu63:
Maybe @kanekosh or @eirikurj can give this a test/review?

kanekosh (approved):
Tested locally with SNOPT and IPOPT and everything worked.

sanjan98 (approved):
Approving since this has already been reviewed by Marco and Shugo and all CI tests passed

PR #434, opened 2025-04-13
Add bare-bones GHA tests

ewu63:
## Purpose

Adds some basic GHA tests, because we do not currently test _optional_ optimizer installation in our docker tests and I want to make sure nothing breaks with 1) optional installation/dependency not breaking, and 2) tests get skipped properly. We should merge this one before #433 to make sure everything looks good with mesonpy.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/build_real.sh
+++ b/.github/build_real.sh
@@ -11,12 +11,7 @@ fi
 mv ~/.config/pip/constraints.txt ~/.config/pip/constraints.txt.bkup
 touch  ~/.config/pip/constraints.txt
 
-# set $PKG_CONFIG_PATH so pkg-config can find ipopt
-# only necessary if IPOPT is installed outside of conda
-export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$IPOPT_DIR/lib/pkgconfig
-
 pip install .[optview,testing] -v
-pip install cyipopt # for testing IPOPT
 
 # move pip constraints file back
 mv ~/.config/pip/constraints.txt.bkup ~/.config/pip/constraints.txt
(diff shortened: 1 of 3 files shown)


ewu63:
PR is ready, but we should probably consider adding `cyipopt` to the deps layer of the Docker containers, beside IPOPT.

.github/build_real.sh
```diff
@@ -16,6 +16,7 @@ touch  ~/.config/pip/constraints.txt
 export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:$IPOPT_DIR/lib/pkgconfig
 
 pip install .[optview,testing] -v
+pip install cyipopt # for testing IPOPT
```
kanekosh, line 19: I'm wondering why we need to install cyipopt as we don't install IPOPT itself and we skip all the IPOPT tests anyway?

kanekosh (changes requested):

.github/build_real.sh
ewu63: This script is used only by the Docker-based testing, not the GHA steps added in this PR. In #425 I added `cyipopt` to the `testing` set of dependencies for pip, here I moved it out of the testing dep into its own line here, such that we can run the test suite without IPOPT installed and those tests should be skipped correctly.
In the future, we should move cyipopt out of this layer and into the deps layer of Docker, so that the Python wrapper for IPOPT is already there if IPOPT is installed.

eirikurj:
@ewu63 yes I think we should install `cyipopt` on the docker images.

eirikurj:
PR for the containers is up https://github.com/mdolab/docker/pull/281.

ewu63:
We can wait for Docker to be updated first, then remove the cyipopt change.

.github/build_real.sh
kanekosh: Right I confused myself with two different setups, thanks!

kanekosh (approved):

Issue #151, opened 2020-08-11
Check handling of scalar funcs output


nwu63:
## Type of issue
What types of issue is it?
_Select the appropriate type(s) that describe this issue_

- Bugfix (non-breaking change which fixes an issue)

## Description
This was a known problem that may have been fixed, but need to check. When `funcs` returns a scalar value, it should be wrapped to be a numpy array for consistency. May also need to check the relevant places within mdolab/multipoint for additional wrapping of `funcs` there.

This issue is somewhat related to #76.

Issue #150, opened 2020-08-11
Make scalar constraints and variables consistent


nwu63:
## Type of issue
What types of issue is it?
_Select the appropriate type(s) that describe this issue_

- Bugfix (non-breaking change which fixes an issue)

## Description
Update them so that a scalar value is just implemented using a numpy arary of size 1. This will unify the code and simplify many things.

PR #436, opened 2025-04-13
Cleanup error class

### LINKED ISSUE #278: Overhaul the `Error` system
# Description of feature
The current setup is not very good---we use the same `Error` class for all exceptions, which means we cannot catch specific errors. Instead, we must resort to parsing the error message as done in #277. We should probably switch to built-in exceptions for most things, and create specific classes for pyOptSparse-specific errors such as not finding the compiled library. We lose the formatted box, but I don't think that's a big deal.

# Potential solution
- switch back to built-in exceptions for most cases, remove the `Error` class
- create specific exception classes as needed for when the built-in exceptions are not sufficient

ewu63:
## Purpose

Closes #278.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_error.py
+++ b/pyoptsparse/pyOpt_error.py
@@ -5,31 +5,6 @@ Holds a simple error handling class for pyOptSparse
 """
 
 
-class Error(Exception):
-    """
-    Format the error message in a box to make it clear this
-    was a expliclty raised exception.
-    """
-
-    def __init__(self, message):
-        msg = "\n+" + "-" * 78 + "+" + "\n" + "| pyOptSparse Error: "
-        i = 21
-        for word in message.split():
-            if len(word) + i + 1 > 78:  # Finish line and start new one
-                msg += " " * (79 - i) + "|\n| " + word + " "
-                i = 2 + len(word) + 1
-            else:
-                msg += word + " "
-                i += len(word) + 1
-        msg += " " * (79 - i) + "|\n" + "+" + "-" * 78 + "+" + "\n"
-        print(msg)
-        self.message = message
-        super().__init__(message)
-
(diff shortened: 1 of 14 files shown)


pyoptsparse/pyOpt_optimization.py
```diff
@@ -1084,13 +1083,13 @@ def processObjtoVec(self, funcs: Dict1DType, scaled: bool = True) -> NumpyType:
             if objKey in funcs:
                 try:
                     f = np.squeeze(funcs[objKey]).item()
-                except ValueError:
-                    raise Error(f"The objective return value, '{objKey}' must be a scalar!")
```
marcomangano, line 1088: What is the nuance of this syntax? I got it before when the two errors were different, is it just for a more explicit handling of the error?

marcomangano (commented):
This  is looking good! Just a couple of comments:

pyoptsparse/pyOpt_error.py
```diff
@@ -5,31 +5,6 @@
 """
 
 
-class Error(Exception):
```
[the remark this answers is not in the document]
ewu63: I can see us adding pyoptsparse-specific error classes here in the future, so I don't mind keeping

pyoptsparse/pyOpt_optimization.py
ewu63: Yeah, the error from numpy is probably less helpful than this specific message. In a later date I would like to clean up shape handling once and for all, see e.g. #150  and #151.

marcomangano (approved):

kanekosh (approved):

PR #437, opened 2025-04-16
Change solution inform into a dataclass

eirikurj:
## Purpose
Realized I had this lying around locally from an earlier development. This PR replaces the standard `dict` with a `dataclass` that gets passed and used to print the optimizer exit code and the associated text. Functionally, there is should be no difference.
Additionally, slipped in one small update to the warning, printing the pyoptsparse versions that's in use and the one that the history was generated with. Can make this as separate PR if needed.

## Expected time until merged
No rush

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


pyoptsparse/pyIPOPT/pyIPOPT.py
```diff
@@ -274,10 +274,8 @@ def intermediate(_, *args, **kwargs):
                 self.hist.writeData("metadata", self.metadata)
                 self.hist.close()
 
-            # Store Results
-            sol_inform = {}
-            sol_inform["value"] = info["status"]
-            sol_inform["text"] = self.informs[info["status"]]
+            # Store optimizer exit condition and reason
+            solInform = SolutionInform(status, self.informs[status])
```
marcomangano, line 278: Looks like `SolutionInform` is not imported in this file

marcomangano (changes requested):
See issue below:

marcomangano (dismissed):

pyoptsparse/pyALPSO/pyALPSO.py
```diff
@@ -191,11 +191,11 @@ def objconfunc(x):
             # Broadcast a -1 to indcate NSGA2 has finished
             self.optProb.comm.bcast(-1, root=0)
 
-            # Store Results
-            sol_inform = {"value": "", "text": ""}
+            # Optimizer has no exit conditions, so nothing to set
+            solInform = None
```
ewu63, line 195: Let's try to stick to pep8 naming for new variables e.g. `sol_inform`

pyoptsparse/pyOpt_solution.py
```diff
@@ -8,8 +10,16 @@
 from .pyOpt_optimization import Optimization
 
 
+@dataclass(frozen=True)
+class SolutionInform:
+    """Data class that contains the optimizer solution value and reason"""
+
+    value: int
+    reason: str
```
ewu63, line 18: Somewhat prefer this field to be called `message` instead of `reason`

ewu63 (commented):
Changes look OK. However, I have a slightly different implementation idea: what if we made `SolutionInform` an `IntEnum` which is created for each optimizer class, and the final outcome is an instance of this class which is the actual exit condition? I think that would lead to a slightly cleaner design. However, in the longer term I would actually prefer _not_ storing this inform mapping ourselves since optimizers may change it up over time and we should not need to maintain it (esp when supporting multiple optimizer versions). So maybe it's not worth it.

marcomangano:
@ewu63 you beat me on time. I followed back the commits and saw you added the `fromInforms` method.

This LGTM, but just a couple of things:
- should we rename the function argument to `sol_inform` as you suggested?
-  very minor, but bit confusing that the `Solution` class in `pyOpt_solution` takes:
`(self, optProb, xStar, fStar, lambdaStar, optInform, info):`
while when the class is instantiated within `_createSolution` in `pyOpt_optimizer` takes:
`(optProb, xStar, fStar, multipliers, sol_inform, info)`
Consistent naming (use `multipliers, sol_inform`) could help future debugging?

ewu63:
I did rename `solInform` to `sol_inform`, but did not touch any class attributes e.g. `optInform`. I would like to, but this goes into bigger problems with backwards compatibility and PEP8 naming conventions so I'd like to punt that to the "near future" if possible. I'd rather do some refactoring and deduplicating code first. Would prefer to also sneak in #402 to the same patch release.

marcomangano:
Sounds good. This and the unconstrained test PR are enough for a patch release. I will look at the other PR later today, we should merge that first since you already bumped the version here.

ewu63 (approved):

kanekosh (approved):

kanekosh:
LGTM! Not merging this yet in case @eirikurj wants to check the changes @ewu63 made.

eirikurj:
Sorry for the absence on this. @ewu63 thanks for updating the PR. Maybe I misunderstood your earlier comment, but was it your intent to change `solInform` to `sol_inform` in this PR (seems like some rebase issue happened)?
Here, the `Solution` class used `optInform`, so I was going for consistency. This variable, however, should not be an issue to change.  I am also fine merging this as-is and refactor later, to avoid delaying further. Just let me know and I can make changes.
I would however like to discuss the pep8 naming further, as this constantly comes up in our codes and we should work based on some written guidelines.  Perhaps we can discuss on this issue https://github.com/mdolab/.github/issues/66 as I think we should try to enforce this.

ewu63:
I guess there was some rebase issues, I renamed the variables but left everything else intact. I agree that we should discuss naming conventions in mdolab/.github#66. The PR is now ready.

PR #441, opened 2025-05-02
Remove invalid escape sequence from docstring

dingraha:
## Purpose
This PR removes an invalid escape sequence from one docstring, and so fixes the warning:

```
> python
Python 3.13.3 | packaged by conda-forge | (main, Apr 14 2025, 20:44:03) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pyoptsparse import pySNOPT
/home/dingraha/projects/aviary_propeller/ved/pyoptsparse/pyoptsparse/pyOpt_optimizer.py:979: SyntaxWarning: invalid escape sequence '\*'
  \*args, \*\*kwargs : varies
>>>
```

Here is a stackoverflow post explaining the issue: https://stackoverflow.com/questions/52335970/how-to-fix-syntaxwarning-invalid-escape-sequence-in-python, and here are the valid escape sequences in Python: https://docs.python.org/3/reference/lexical_analysis.html#escape-sequences. You'll see that `\*` is not listed.

## Expected time until merged
Not urgent.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
No testing needed, but I did run the tests in `tests` and they seemed OK (but I don't have all the optimizers pyoptsparse supports so it's tough to tell).

## Checklist

### DIFF
--- a/pyoptsparse/pyOpt_optimizer.py
+++ b/pyoptsparse/pyOpt_optimizer.py
@@ -976,7 +976,7 @@ def OPT(optName, *args, **kwargs):
        Either a string identifying the optimizer to create, e.g. "SNOPT", or
        an enum accessed via ``pyoptsparse.Optimizers``, e.g. ``Optimizers.SNOPT``.
 
-    \*args, \*\*kwargs : varies
+    *args, **kwargs : varies
        Passed to optimizer creation.
 
     Returns


kanekosh (approved):
Thank you for catching this!

ewu63:
Not sure if this was the intention
image

What I would do instead is either double backslash e.g. `\\*` as seen [here](https://github.com/mdolab/pygeo/blob/e2cdcbd567d4ebe4f22f64281a23da4dc0fa7a3c/pygeo/parameterization/DVGeoCST.py#L321), or single backslash as we had before `\*` but wrap the entire docstring as a raw string, i.e. `r"""...`.

Issue #442, opened 2025-05-13
`numpy>=2.0` requirement in pyproject.toml


zfrederi:
## Description
The numpy requirement differs between the `setup.py` and `pyproject.toml` files. In `setup.py` it is 
`numpy>=1.21`

whereas in `pyproject.toml` it is 
`numpy>=2.0`

### Steps to reproduce issue
When trying to install pyoptsparse, I was prompted to update to numpy 2 which I don't think is an actual requirement to run pyoptsparse. 
1. `conda install pyoptsparse`

### Potential solution
Change the numpy requirement in `pyproject.toml` to be the same as `setup.py`.

ewu63:
To clarify, `setup.py` contains the run requirement (numpy 1 or 2) and `pyproject.toml` has `numpy>=2.0` in the build-systems section, which is the build requirement. Both of these should be correct.

Did you use conda to install pyoptsparse? The dependency management in conda is independent of those mechanisms above, and is reliant on the information within the conda-forge feedstock. The relevant lines are [here](https://github.com/conda-forge/pyoptsparse-feedstock/blob/main/recipe/meta.yaml#L29) and does not in fact pin the numpy version. Are you sure you're not able to get a numpy1 environment? The following generates a valid conda environment for me:
```
conda create --name test pyoptsparse numpy==1.26
```

ewu63:
Based on discussion in #443 this may be a conda issue unrelated to pyoptsparse.

PR #443, opened 2025-05-21
Change minimum numpy version in pyproject to match setup

jkirk5:
## Purpose
`setup.py` specifies "numpy>1.21", but `pyproject.toml` restricts "numpy>2". It's possible this was an oversight, since pyoptsparse does not strictly require numpy 2, and forcing this version breaks compatibility with any other packages that don't support numpy 2 yet. I've set the required numpy version in `pyproject.toml` to match ">1.2.1"

This should hopefully cover everyone satisfactorily - you will get numpy>2 by default (installing pyoptsparse first), but if your environment already has a numpy 1.X version then this change should allow you to continue using it without forcing an upgrade.

## Expected time until merged
1 week

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [x] Other (please describe)

Allows for use of latest pyoptsparse with environments that require numpy<2 (so not a "bug" with pyoptsparse itself)

## Testing
In fresh Python/conda env, install numpy=1.26
Install pyoptsparse (try with both pip and conda installs)
- List of new packages to be installed should NOT include numpy>=2
`conda list` and `pip list` should still point to numpy=1.26
Activating Python and running:
```
import numpy as np
foo = np.Inf
```
should NOT return an error - this function will raise an error if numpy>=2 was still added to the environment (I've seen conda sometimes not show in `conda list` that a package version was updated)

## Checklist
- [x] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable

As far as I'm aware there is nowhere in the documentation that talks about using numpy>2 so no changes required

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=42", "meson>=0.60.0", "ninja", "numpy>=2.0", "swig"]
+requires = ["setuptools>=42", "meson>=0.60.0", "ninja", "numpy", "swig"]
 build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=42", "meson>=0.60.0", "ninja", "numpy", "swig"]
+requires = ["setuptools>=42", "meson>=0.60.0", "ninja", "numpy>=1.21", "swig"]
 build-backend = "setuptools.build_meta"


jkirk5:
> I'm not able to reproduce the issue. With a clean environment and only `numpy 1.26` installed, if I run `pip install .` within the pyoptsparse source directory, I do not get a new version of numpy installed. Similarly if I do `conda install pyoptsparse` I do not see any upgrade on numpy. As I explained in [this comment](https://github.com/mdolab/pyoptsparse/issues/442#issuecomment-2878241693) the version specification in `pyproject.toml` is for the build system only, and does not affect runtime requirements (which is still numpy 1 or 2).

Ah sorry for not seeing that existing issue! My environment has the same problem where numpy was already pip installed, so perhaps this is a conda problem. I will try reversing installation order of my packages and see if that helps, and will close the PR if it does!

PR #455, opened 2025-08-22
Ruff fixes

A-CGray:
## Purpose

Updates repo to pass our new formatting and linting checks. There are a lot of whitespace changes here so once this PR is merged we can add the commit hash to a `.git-blame-ignore-revs` file. 

## Type of change

- [x] Code style update (formatting, renaming)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


ewu63 (approved):

ewu63:
I really don't think we should be modifying the optimizer source code which we just vendor in - this large diff doesn't do anything for us.

A-CGray:
> I really don't think we should be modifying the optimizer source code which we just vendor in - this large diff doesn't do anything for us.

I know it's a big diff, but we can put this PR's commit in `.git-blame-ignore-revs`. The kind of trailing whitespace fixes that are being applied to those source files are ones that many IDEs apply automatically, so doing this one off fix here avoids people unwittingly making these changes in the future.

> I'm also a little confused. How is the pre-commit supposed to be applied outside of CI? And is there a way to configure those per-repo? If the answer is no to either of those then we may want to rethink the pre-commit approach.

Download the config file from the .github repo and put it in the repo in question (or put it somewhere central and softlink to it in the repo). `pip install pre-commit`, then you can manually run the checks with `pre-commit run --all-files`, or set them up to run on commits with `pre-commit install`. Extending pre-commit configs is [explicitly not supported](https://github.com/pre-commit/pre-commit/issues/1203).

ewu63:
I think ultimately, my gripe is with the lack of configurability with this current setup. In the past, we had allowed each repo to define the set of rules that they would be compliant with (black, flake8, isort, pylint, etc.). Now, each repo is required to be compliant to all 10 rules stored [here](https://raw.githubusercontent.com/mdolab/.github/main/.pre-commit-config.yaml), and I think that causes unnecessary issues. Again, I think the vast majority of changes in this PR are somewhat unnecessary and difficult to review thoroughly. In addition to ruff, I'm fine with the mixed line ending changes.

In any case, I don't want this PR to block all development in pyOptSparse, I just don't think a `+1304, -1310` PR of mostly whitespace changes is helpful.

ewu63 (approved):

kanekosh (approved):

PR #466, opened 2025-11-16
Update types

ewu63:
## Purpose
Just some long-overdue type annotations

## Expected time until merged
A few days

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


marcomangano:
The error on readthedocs must be a glitch (`setuptools` missing from the environment?) because the other PRs are not having the same issue.

The stable build is failing because that image is still using python<3.10, so there is a version conflict. Let me try real quick to move back the requirement to see if at least the RTD build passes

ewu63:
Finally ready @marcomangano @A-CGray

pyoptsparse/pyOpt_gradient.py
```diff
@@ -58,7 +55,7 @@ def __init__(self, optProb: Optimization, sensType: str, sensStep: float = None,
         else:
             self.mydvs = list(range(ndvs))
 
-    def _eval_func(self, x: ndarray) -> Tuple[ndarray, ndarray, bool]:
+    def _eval_func(self, x: ndarray) -> tuple[ndarray, ndarray, bool]:
```
marcomangano: Should we use the numpy-specific array type here as well?

marcomangano (commented):
I gave a quick pass and it is looking good. This is only tested locally, correct? Do we have any notes about that?

A-CGray:
How do you want to handle this PR  relative to #483 @ewu63? Merge that first then add typing for the Uno wrapper in this PR?

pyoptsparse/pyOpt_gradient.py
ewu63: Done

ewu63:
This PR is ready for review @marcomangano @A-CGray

A-CGray:
@ewu63 do we want to add type checking to CI? We do have [this mypy action](https://github.com/mdolab/.github/blob/main/.github/workflows/mypy.yaml) we could run

pyoptsparse/pyOpt_optimization.py
A-CGray: If this branch is taken the return type doesn't match the annotation (array instead of dict)

ewu63:
Eventually yes but probably not now - it definitely won't pass and I would like to do some refactoring/updates before fully sorting out the types. This was just some quick work I had stashed that I wanted merged before doing more work on the code.

A-CGray (changes requested):

A-CGray:
> Eventually yes but probably not now - it definitely won't pass and I would like to do some refactoring/updates before fully sorting out the types. This was just some quick work I had stashed that I wanted merged before doing more work on the code.

Are there any files that you think will pass in current form? We could enable the check and limit it to those files for now, and add more of the codebase as we improve the typing coverage. That way we will at least have some protection against degradation.

pyoptsparse/pyOpt_optimization.py
```diff
@@ -78,7 +80,7 @@ def __init__(self, name: str, objFun: Callable, comm=None, sens: Optional[Union[
         self.invXScale: ndarray = None
         self.xOffset: ndarray = None
```
A-CGray: Replace `ndarray` with `npt.NDArray[np.floating]` or `ArrayType`? Then you can remove `from numpy import ndarray`

ewu63:
> Are there any files that you think will pass in current form? We could enable the check and limit it to those files for now, and add more of the codebase as we improve the typing coverage. That way we will at least have some protection against degradation.

I doubt it, and setting up mypy correctly will take time -- in the future I am interested in using something like `pyrefly` for type checking, but this codebase is nowhere ready for type checking in CI.

pyoptsparse/pyOpt_optimization.py
ewu63: Done

pyoptsparse/pyOpt_optimization.py
ewu63: Fixed

A-CGray (approved):

marcomangano (approved):

PR #478, opened 2026-03-10
Slow down TP109 function evaluation to fix SNOPT testing issue

A-CGray:
## Purpose

Fixes #477, where SNOPT was occasionally succesfully solving the TP109 problem despite being given a time limit of 1e-15 seconds.

SNOPT uses the `system_clock` function to measure the elapsed time at each function evaluation. Depending on the integer types passed to this function, [the resulting time measurement may only have millisecond precision](https://gcc.gnu.org/onlinedocs/gfortran/SYSTEM_005fCLOCK.html). I therefore believe the issue is that when function evaluations are on the order of <=1ms and the default fortran integer is 32 bit (kind=4), SNOPT will often register function evaluations as taking zero time, thus messing up the enforcement of the time limit.

## Expected time until merged

Soon plz

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/tests/test_tp109.py
+++ b/tests/test_tp109.py
@@ -30,6 +30,7 @@ Test uses Schittkowski's TP109 constraint problem.
 
 # Standard Python modules
 import unittest
+import time
 
 # External modules
 import numpy as np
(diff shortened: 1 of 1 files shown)


ewu63 (approved):
Makes sense and I had a suspicion also that SNOPT was not measuring time to the precision we are after here. Thanks!

PR #483, opened 2026-03-20
Add an unopy interface

### LINKED ISSUE #318: Integrate Uno into pyOptSparse
Integrate Uno into https://github.com/mdolab/pyoptsparse.

cc @ewu63

robfalck:
## Purpose

This pull request adds an interface to [Uno](https://github.com/cvanaret/Uno) to pyoptsparse.
Uno is a C++ package for nonlinear constrained optimization, with a python interface (unopy).
Uno provides another option that provides both SQP and interior-point capability depending on the settings.

Other notes about the Uno interface
- Supports sparse jacobians
- Supports user termination
- Output can be sent to a stream specified by the user using 'logger_stream' option, with verbosity set by the 'logger' option.

This implementation requires unopy 0.4.7 or later.

## Expected time until merged

A few weeks (not urgent).

## Type of change

- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Running the test suite will test Uno against several existing test problems.

## Checklist

Resolves cvanaret/Uno#318


robfalck:
@cvanaret I would appreciate your review again.  Theres a skipped test of problem 109 (TP109) from the Schittkowski test that I wasn't able to get to converge with Uno.

tests/test_tp109.py
```diff
@@ -205,11 +196,13 @@ def test_uno_informs(self):
                             optOptions={"time_limit": 1e-15, "logger": "INFO"})
```
cvanaret: Is `"time_limit": 1e-15` correct?

cvanaret:
@robfalck thanks for the PR!

I have an AMPL model of `hs109` and Uno indeed doesn't converge with `hessian_model="LBFGS"` + the default `quasi_newton_memory_size=6`. It looks like the Hessian approximation is too positive definite and the solver takes tiny steps.
It does converge for `quasi_newton_memory_size=15` but that's just lucky.

I think I should monitor progress and flush the limited memory if the solver stalls. Note that I'm also planning to implement interior points with L-BFGS in the next few days.

Will the exact Hessian callback be available in the near future?

robfalck:
Thanks for checking.  That last question is up to the mdolab folks since it's a significant change.

tests/test_tp109.py
robfalck: That was a test to make sure the solution inform indicates that the time limit was reached.

cvanaret:
[unopy v0.4.1](https://pypi.org/project/unopy/0.4.1/) is out with:
- L-BFGS for interior-point methods:

```python
uno_solver.set_preset("ipopt")
# either set the Hessian model explicitly
uno_solver.set_option("hessian_model", "LBFGS")
# or let Uno default to L-BFGS if no Hessian is provided
```
- a function `set_logger_stream`:

```python
uno_solver.set_logger_stream(sys.stdout)
uno_solver.optimize(model)
sys.stdout.flush()

# or

with open("log.txt", "w") as f:
    uno_solver.set_logger_stream(f)
    uno_solver.optimize(model)
    f.flush()
```

marcomangano:
> Will the exact Hessian callback be available in the near future?

It would be a very cool capability to add since most gradient-based optimizers should support that, but I don't believe that the lab (pinging @A-CGray @eirikurj ) is interested in adding this capability in the near term. Most of the (high-fidelity) codes used for engineering MDAO do not provide second derivatives.

I haven't thought this through yet but I think it would require a change in API of the [`_masterFunc` and `_masterFunc2` callback function wrappers](https://github.com/mdolab/pyoptsparse/blob/5c3696478cff64b54d7c34d0c3c9f46c1f0f5335/pyoptsparse/pyOpt_optimizer.py#L202-L637) and some non-naive way to store sparsity information, so it would not be a trivial implementation. I also know that @ewu63 has been considering some major code refactoring at some point, so this change could fall under that bigger effort.

A-CGray:
> > Will the exact Hessian callback be available in the near future?
> 
> It would be a very cool capability to add since most gradient-based optimizers should support that, but I don't believe that the lab (pinging @A-CGray @eirikurj ) is interested in adding this capability in the near term. Most of the (high-fidelity) codes used for engineering MDAO do not provide second derivatives.

I wouldn't complain if somebody else implemented it, but we have no plans to. We are rarely (if ever) solving problems where the Hessian is computable in any practical sense. If somebody was going to implement this, we may as well implement the ability to use jacobian-vector products as well.

Thanks for doing this work @robfalck , I have been meaning to try out Uno for a long time, now I might actually get around to it!

marcomangano:
Hi all, I am working through some testing environment issues on my side and I am currently running stuff on docker to prepare the other PR. Anyway, I kept bumping into this weird error that makes all the Uno tests fail when running `testflo`:

```TypeError: protected_actual_reduction_macheps_coefficient is not of type int``` triggered on [this line](https://github.com/robfalck/pyoptsparse/blob/812a5ff83c87ec6db5891e9cdf118515d4436aa4/pyoptsparse/pyUno/pyUno.py#L372)

Even if I make sure that the option defined in [this line](https://github.com/robfalck/pyoptsparse/blob/812a5ff83c87ec6db5891e9cdf118515d4436aa4/pyoptsparse/pyUno/pyUno.py#L131) is actually an int, the error persists. Only if I remove the option altogether the tests actually converge. @robfalck do you  have any issues like that on your side?

**Comment by cvanaret** [comment]
Hi @marcomangano, can you try `10.` instead of `10`?
The option should be a float but it looks like the wrong unopy overload is called.

**Comment by marcomangano** [comment]
> Hi @marcomangano, can you try `10.` instead of `10`? The option should be a float but it looks like the wrong unopy overload is called.

Ha, that fixes it, good catch! I am not very fluent in cpp, is it because `10` is read as an int by the cpp layer [here](https://github.com/cvanaret/Uno/blob/45eb9929e0f55e70a0c9dbb77d66fc3c21c8227c/interfaces/Python/python_modules/UnoSolver.cpp#L38)? ~~Anyway, pushing the quick fix here for now.~~ I cannot commit directly to Rob's branch, but that is an easy fix for later.

**Comment by cvanaret** [comment]
> Ha, that fixes it, good catch! I am not very fluent in cpp, is it because `10` is read as an int by the cpp layer [here](https://github.com/cvanaret/Uno/blob/45eb9929e0f55e70a0c9dbb77d66fc3c21c8227c/interfaces/Python/python_modules/UnoSolver.cpp#L38)?

Great!
Yes, it picks the right function given the type of the argument, so with `10` it will naturally pick the integer one. A bit stringent, but this avoids implicit conversions between types.

**Comment by robfalck** [comment]
Don't have access to a computer but hopefully my update today via GitHub mobile fixed it.

**Comment by cvanaret** [comment]
[unopy v0.4.2](https://pypi.org/project/unopy/0.4.2/) is out. The `ipopt` preset with L-BFGS Hessian should be able to solve `hs109`.

**Comment by robfalck** [comment]
Setting the log stream via the `logger_stream` option works now.

**Comment by A-CGray** [comment]
FYI, we've merged an update to our docker images so that they include Uno, the images are rebuilding at the moment so should be available in a few hours.

I found that Uno takes a really long time on the Rosenbrock test because it can't reduce the optimality quite enough to reach it's tolerance and so runs right up to it's iteration limit. The tests still pass though because it does find a solution close enough to the true solution. Maybe we want to switch to the `ipopt` preset for that test?

**Comment by cvanaret** [comment]
@A-CGray interesting! Can you point me to the model, so that I can fiddle with it? Perhaps L-BFGS can be better initialized.

**Comment by A-CGray** [comment]
> @A-CGray interesting! Can you point me to the model, so that I can fiddle with it? Perhaps L-BFGS can be better initialized.

It's [this test](https://github.com/robfalck/pyoptsparse/blob/997a063558d75405e877053d964c839c4c7ca07b/tests/test_rosenbrock.py#L147-L158). For debugging Uno, I commented out the other optimizers in the `parameterized_expand` and the line that sets the logger to `"SILENT"`. It seems that by default Uno was using SR1, not BFGS, this was with Unopy 0.4.4. The point it finds is for all practical purposes the optimum (the objective is ~2e-11, whereas the true optimum is 0), it just can't converge those last few orders of magnitude and so spends ~200 iterations bouncing around this point.

**Comment by cvanaret** [comment]
@A-CGray I was able to reproduce. I would suggest these default options be set in `pyUno.py`:
```python
"hessian_model": [str, "LBFGS"], # avoid Uno defaulting to L-SR1 for trust-region methods
"quasi_newton_memory_size": [int, 20],
```

cvanaret:
Did the change break other tests? I'm not sure what `self.assertTrue(log_file_size > prev_log_file_size)` refers to.

robfalck:
Yeah checking it now. I was trying to ensure that a log file longer under more verbose output, short of testing content from each individual file

robfalck:
I think the change reduced the warnings, such that the file size of the WARNING logger was the same as the file size of the DISCRETE logger.

cvanaret:
Perfect. Let me know if you need anything.

tests/test_hs071.py
```diff
@@ -198,17 +200,47 @@ def test_ipopt_informs(self):
         sol = self.optimize(optOptions={"max_cpu_time": 0.001})
         self.assert_inform_equal(sol, -4)
 
+    def test_uno_informs(self):
+        self.optName = "Uno"
+        self.setup_optProb()
+        # Test that the inform is 1 (Iteration Limit Exceeded) when iterations are too limited.
+        sol = self.optimize(optOptions={"max_iterations": 1, "logger": "SILENT"})
+        print(sol)
```
marcomangano: Is this a debugging leftover or should we keep it to have a more verbose output?

doc/optimizers/UNO_options.yaml
```diff
@@ -0,0 +1,491 @@
+preset:
+  desc: |
+    Specifies the preset option in Uno (str).
+
+    - ``filtersqp``: Trust region restoration filter SQP solver.
+    - ``filterslp``: Trust region restoration filter SLP solver.
+    - ``funnelsqp``: Funnel-method SQP solver.
+    - ``ipopt``: line-search filter restoration infeasible interior-point solver
+
+    IPOPT requires a direct linear solver and is not currently functional with pyOptSParse.
```
cvanaret:
```suggestion
    The `ipopt` preset requires a direct linear solver and is not currently functional with pyOptSParse.
```

marcomangano: We should remove this altogether as we actually can run the `ipopt` preset now!

marcomangano:
> I think the change reduced the warnings, such that the file size of the WARNING logger was the same as the file size of the DISCRETE logger.

Looks like the file size is a bit inconsistent here, somehow now the `WARNING` logger is smaller than `DISCRETE` on that specific intel image. Any ideas?

EDIT: It looks like it is happening inconsistently, not sure if/how to address it. Could be annoying if the whole testing stack (to deploy the Docker images) fails randomly because of this.

A-CGray:
> > I think the change reduced the warnings, such that the file size of the WARNING logger was the same as the file size of the DISCRETE logger.
> 
> Looks like the file size is a bit inconsistent here, somehow now the `WARNING` logger is smaller than `DISCRETE` on that specific intel image. Any ideas?
> 
> EDIT: It looks like it is happening inconsistently, not sure if/how to address it. Could be annoying if the whole testing stack (to deploy the Docker images) fails randomly because of this.

Is it a given that the `WARNING` log setting should always produce as much or more output than the `DISCRETE` log setting? If not then checking the file size is probably too flaky a thing to test (other than for the `SILENT` option). Are there some distinct strings we expect to see in the log for each level that we could check for instead?

robfalck:
I'll just test for the presence of files.  I was trying to verify that we were getting expected behavior but at that point it's probably more of an Uno test than a pyoptsparse one

tests/test_hs071.py
marcomangano: @robfalck this is my only leftover comment, the PR looks ready to go for me!

robfalck: print removed, thanks

robfalck:
@marcomangano I think I've resolved the last issue you had with this?  Is there anything else holding it up that I can address?

We'll be releasing the next version of OpenMDAO soon and I would like to be able to update our pyoptsparse driver to be able to utilize this.

marcomangano (dismissed):
Sorry for the delay, I have been afk for a couple of days. LGTM!

@A-CGray should we make a new (minor) release?

A-CGray:
> @A-CGray should we make a new (minor) release?

Yeah this is a fairly major new functionality so definitely a minor release.

@robfalck how imminent is this new OpenMDAO release? Given that this is quite a major new feature I'd like to have a review from one more pyOptSparse maintainer. I should have time to look through this at the start of next week, or @ewu63 might be available sooner?

robfalck:
Next week is fine.  I'll get the Uno settings into the pyoptsparse driver on the OpenMDAO side.  We'll probably be releasing in a week or two

pyoptsparse/pyUno/pyUno.py
```diff
@@ -0,0 +1,379 @@
...
+        storeSens : bool
+            Flag specifying if sensitivities are to be stored in hist.
+            This is necessary for hot-starting only.
+        """
+        self.startTime = time.time()
+        self.callCounter = 0
+        self.storeSens = storeSens
+
+        if len(optProb.constraints) == 0:
+            # Uno requires at least one constraint, so add a dummy one
```
ewu63: Is this true, that unopy requires a dummy constraint and cannot handle unconstrained problems by itself?

cvanaret: I have no idea what the code refers to 😅

ewu63: Yeah I suspect it's just a copy-paste from something else, can probably remove and just let Uno handle it.

robfalck: My apologies, I had misunderstood the intent of dummyConstraint. Errant comment removed.

marcomangano (dismissed):
Good catch @ewu63

pyoptsparse/pyUno/pyUno.py
```diff
@@ -0,0 +1,377 @@
...
+        if isinstance(unopy, Exception) and raiseError:
+            raise unopy
+
+        unopy_version = ilmd.version("unopy")
+
+        if Version(unopy_version) < Version(_UNOPY_MIN_VERSION):
+            raise RuntimeError(
+                "The pyoptsparse Uno interface requires unopy "
+                f"{_UNOPY_MIN_VERSION} or later, but {unopy_version} is installed"
+            )
```
A-CGray: If `raiseError=False` and unopy isn't installed, won't this raise an error when it tries to run `unopy_version = ilmd.version("unopy")`? Surely we only want to get the unopy version if it's actually installed?

```suggestion
        if isinstance(unopy, Exception):
            if raiseError:
                raise unopy
        else:
            unopy_version = ilmd.version("unopy")
            if Version(unopy_version) < Version(_UNOPY_MIN_VERSION):
                raise RuntimeError(
                    "The pyoptsparse Uno interface requires unopy "
                    f"{_UNOPY_MIN_VERSION} or later, but {unopy_version} is installed"
                )
```

pyoptsparse/pyUno/pyUno.py
```diff
@@ -0,0 +1,377 @@
...
+            IDENTICAL**. As soon as the requested evaluation point does
+            not match the history, function and gradient evaluations
+            revert back to normal evaluations.
+
+        storeSens : bool
+            Flag specifying if sensitivities are to be stored in hist.
+            This is necessary for hot-starting only.
+        """
+        self.startTime = time.time()
+        self.callCounter = 0
```
A-CGray, line 213: Need to reset `self._userRequestedTermination = False` otherwise one user-requested termination will terminate all future invocations of `__call__`

```suggestion
        self.callCounter = 0
        self._userRequestedTermination = False
```

pyoptsparse/pyUno/pyUno.py
```diff
@@ -0,0 +1,377 @@
...
+
+            solver = unopy.UnoSolver()
+            self._set_uno_options(solver)
+
+            try:
+                self.result = solver.optimize(model)
+            except KeyboardInterrupt:
+                # User requested termination during optimization
+                # Create a result object indicating user termination
+                pass
```
A-CGray: Is this intended to work for actual keyboard interrupts (i.e when the user ctrl+c's while the optimiser is running)? If so then we should set `self._userRequestedTermination=True` in here so that the inform is correctly set below, otherwise you'll get an error when trying to access `self.result.optimization_status.value`.

If this is not intended to handle a real keyboard interrupt then we should use a different a different (potentially custom?) error type to handle a user requested termination

doc/optimizers/UNO_options.yaml
```diff
@@ -0,0 +1,489 @@
...
+    - ``SILENT``, ``DISCRETE``, ``WARNING``, ``INFO``, ``DEBUG``, ``DEBUG2``, ``DEBUG3``
+
+    The default value is ``INFO``.
+
+logger_stream:
+  desc: |
+    The stream to which the logger writes the log.
+
+    The default value is None, which leaves the logger stream unchanged.
+
```
A-CGray, line 99: How much effort would it be to allow the user to pass in a filename as a string and have pyoptsparse/Uno handle opening the file?

A-CGray (changes requested):

pyoptsparse/pyUno/pyUno.py
robfalck: Good point. I'll implement a custom _UserTerminationException so that we can handle those explicitly and not accidentally swallow a RuntimeError or ValueError that might get raised.

doc/optimizers/UNO_options.yaml
robfalck: That can be handled on the pyoptsparse side pretty easily. If the logger_stream option is a string or a path, we'll open it, and close it in a try/finally block wrapped around `solver.optimize`.

A-CGray: That would be great to support, because then if running in MPI I can just pass in a string, and pyOptSparse can make sure to only open the file on the root proc, rather than me having to worry about doing that.

A-CGray (dismissed):

robfalck:
I think we discussed making a conda-recipe for uno/unopy to make it a more easily-installed depdendency.  @ewu63 have you had a chance to start that?

A-CGray:
Thanks for addressing those comments  @robfalck . There is one last thing I just remembered about that I'd like to include here, which is marking certain iterations as "major" in the history file, we may need @cvanaret 's help for this.

Looking at Uno's log, it seems like it already makes this distinction between major and minor evaluations, if my understanding is correct (@cvanaret ) then the evaluations we'd want to mark as major are the ones with a  ✔ in their status. These are the points from which you begin your linesearch/trust region search in the next iteration right?

```
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  Iterations
 Major  Minor  Penalty   Radius    Phase Funnel    |BFGS|  ||Step||  Objective  Infeas    Statio    Compl     Status
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 0      -      -         1.00e+01  OPT   1.00e+00  0       -         1.117e+00  5.43e-01  2.33e-02  0.00e+00  initial point
 1      1      -         1.00e+01  OPT   7.85e-01  0       6.67e-02  1.087e+00  5.70e-01  7.33e-02  6.31e-03  ✔ (h-type)
 2      1      -         1.00e+01  OPT   7.85e-01  1       6.81e-02  1.020e+00  9.76e-01  -         -         ✘ (funnel)
 -      2      -         3.41e-02  OPT   -         -       -         -          -         -         -         infeasible
 -      2      0.00e+00  3.41e-02  FEAS  -         0       3.41e-02  1.092e+00  6.50e-01  -         -         ✘ (Armijo)
 -      3      0.00e+00  1.70e-02  FEAS  -         1       1.70e-02  1.090e+00  5.21e-01  7.99e+00  2.15e-01  ✔ (Armijo)
 3      1      0.00e+00  3.41e-02  FEAS  -         2       3.41e-02  1.096e+00  4.68e-01  1.00e+00  2.08e-01  ✔ (Armijo)
 4      1      0.00e+00  6.81e-02  FEAS  -         3       6.81e-02  1.107e+00  3.81e-01  2.80e+00  2.99e-01  ✔ (Armijo)
 5      1      0.00e+00  1.36e-01  FEAS  -         4       1.10e-01  1.118e+00  3.60e-01  1.38e+00  1.68e-01  ✔ (Armijo)
 6      1      0.00e+00  1.36e-01  FEAS  -         5       1.01e-01  1.129e+00  2.45e-01  8.63e-01  8.06e-02  ✔ (Armijo)
 7      1      0.00e+00  1.36e-01  FEAS  -         6       7.70e-02  1.138e+00  1.53e-01  4.71e-01  5.02e-02  ✔ (Armijo)
 8      1      0.00e+00  1.36e-01  FEAS  -         7       6.86e-02  1.146e+00  9.69e-02  8.75e-01  4.39e-02  ✔ (Armijo)
 9      1      0.00e+00  1.36e-01  FEAS  -         8       4.35e-02  1.151e+00  4.39e-02  2.69e-01  6.75e-03  ✔ (Armijo)
 10     1      0.00e+00  1.36e-01  FEAS  -         9       2.26e-02  1.153e+00  9.67e-03  1.22e-01  5.46e-04  ✔ (Armijo)
 11     1      0.00e+00  1.36e-01  FEAS  -         10      6.06e-03  1.153e+00  9.85e-04  2.84e-02  3.71e-06  ✔ (Armijo)
 12     1      -         1.36e-01  OPT   3.97e-01  2       3.59e-02  1.106e+00  1.27e-01  1.21e+00  3.23e-02  ✔ (f-type)
 13     1      -         1.36e-01  OPT   3.97e-01  3       4.13e-02  1.061e+00  2.29e-01  1.32e+00  4.42e-02  ✔ (f-type)
 14     1      -         1.36e-01  OPT   3.97e-01  4       4.21e-02  1.025e+00  5.25e-01  -         -         ✘ (funnel)
 -      2      -         2.11e-02  OPT   -         -       -         -          -         -         -         infeasible
 -      2      0.00e+00  2.11e-02  FEAS  -         11      2.11e-02  1.066e+00  1.65e-01  2.08e+00  4.97e-02  ✔ (Armijo)
 15     1      0.00e+00  4.21e-02  FEAS  -         12      2.15e-02  1.069e+00  9.42e-02  1.81e-01  4.43e-03  ✔ (Armijo)
 16     1      -         4.21e-02  OPT   2.81e-01  5       2.77e-02  1.038e+00  2.22e-01  5.38e-01  2.14e-02  ✔ (f-type)
 17     1      -         4.21e-02  OPT   2.81e-01  6       4.21e-02  1.004e+00  5.73e-01  -         -         ✘ (funnel)
 -      2      -         2.11e-02  OPT   -         -       -         -          -         -         -         infeasible
 -      2      0.00e+00  2.11e-02  FEAS  -         13      2.11e-02  1.048e+00  1.04e-01  5.32e+00  5.52e-02  ✔ (Armijo)
 ```
 
The question is, how can we tell whether a given evaluation is major or minor from within pyoptsparse? If you look at [SNOPT](https://github.com/mdolab/pyoptsparse/blob/769adbc8344d939550d535a7766ba340c815e90b/pyoptsparse/pySNOPT/pySNOPT.py#L629) and [IPOPT](https://github.com/mdolab/pyoptsparse/blob/769adbc8344d939550d535a7766ba340c815e90b/pyoptsparse/pyIPOPT/pyIPOPT.py#L257), they both set the `isMajor` flag to `True` inside callback functions from their respective optimisers that are known to be called only during major iterations. Is there something similar in Uno?

**Comment by A-CGray** [comment]
> I think we discussed making a conda-recipe for uno/unopy to make it a more easily-installed depdendency. @ewu63 have you had a chance to start that?

In the end we realised that `unopy` includes the uno binary (we initially though it was just the python wrapper) so, on linux at least, installing uno is as simple as `pip install unopy`. No idea if that works as smoothly on Mac and Windows, if not then I guess we'd still want a conda install method.

**Comment by marcomangano** [comment]
> I think we discussed making a conda-recipe for uno/unopy to make it a more easily-installed depdendency. @ewu63 have you had a chance to start that?

FYI, we ended up just installing `unopy` with `pip` in our environment and it works out of the box, so maybe the work on the conda package is not as urgent anymore?
Some of the early confusion / installation issues came from the fact that I thought `unopy` was just a wrapper like `cyipopt`, but it actually comes with all the Uno binaries :sweat_smile:.

EDIT: Ali beat me for a few seconds! I agree that the Windows compatibility might be a good reason for looking into Conda, but I think it should be separate from this PR / initial implementation.

**Comment by cvanaret** [comment]
Our CI workflow successfully builds and tests Python wheels on macOS and Windows too :) See the available wheels on [PyPI](https://pypi.org/project/unopy/#files).

**Comment by cvanaret** [comment]
> Thanks for addressing those comments @robfalck . There is one last thing I just remembered about that I'd like to include here, which is marking certain iterations as "major" in the history file, we may need @cvanaret 's help for this.
> 
> Looking at Uno's log, it seems like it already makes this distinction between major and minor evaluations, if my understanding is correct (@cvanaret ) then the evaluations we'd want to mark as major are the ones with a ✔ in their status. These are the points from which you begin your linesearch/trust region search in the next iteration right?
> 
> The question is, how can we tell whether a given evaluation is major or minor from within pyoptsparse? If you look at [SNOPT](https://github.com/mdolab/pyoptsparse/blob/769adbc8344d939550d535a7766ba340c815e90b/pyoptsparse/pySNOPT/pySNOPT.py#L629) and [IPOPT](https://github.com/mdolab/pyoptsparse/blob/769adbc8344d939550d535a7766ba340c815e90b/pyoptsparse/pyIPOPT/pyIPOPT.py#L257), they both set the `isMajor` flag to `True` inside callback functions from their respective optimisers that are known to be called only during major iterations. Is there something similar in Uno?

Good question. Indeed only the major iterations are marked with ✔. The inner iterations are line-search or trust-region iterations during which trial iterates are rejected.
We have [such a callback in Uno's C API](https://github.com/cvanaret/Uno/blob/main/interfaces/C/Uno_C_API.h#L127), but I haven't implemented it in unopy yet. Should I quickly add it?

**Comment by marcomangano** [comment]
> Good question. Indeed only the major iterations are marked with ✔. The inner iterations are line-search or trust-region iterations during which trial iterates are rejected. We have [such a callback in Uno's C API](https://github.com/cvanaret/Uno/blob/main/interfaces/C/Uno_C_API.h#L127), but I haven't implemented it in unopy yet. Should I quickly add it?

It would be great if you can add the feature and make a new unopy release so we can bump the environment and add a test. That is a good catch by @A-CGray, tagging the major iterations is very helpful for debugging, restarting the optimization, etc.

**Comment by cvanaret** [comment]
The callback is available in [unopy 0.4.7](https://pypi.org/project/unopy/0.4.7/), see the [example file](https://github.com/cvanaret/Uno/blob/main/interfaces/Python/example/example_hs015.py#L40).

**Comment by A-CGray** [comment]
Thanks @cvanaret , I've been working on getting that major iteration tracking working with that callback. I have it working but there seems to have been some change between unopy v0.4.6 and 0.4.7 that means Uno can no longer converge the rosenbrock test problem. With 0.4.6 it converges in 47  major iterations, whereas with 0.4.7 I killed the process after about 1500 major iterations without convergence. I've attached two log files from the two versions, where the 0.4.7 run was limited to 100 major iterations, I don't see any obvious differences in the algorithm settings.

[Uno 046-rosenbrockOutput.log](https://github.com/user-attachments/files/27781294/Uno.046-rosenbrockOutput.log)
[Uno 047-rosenbrockOutput.log](https://github.com/user-attachments/files/27781295/Uno.047-rosenbrockOutput.log)

**Comment by ewu63** [comment]
> I think we discussed making a conda-recipe for uno/unopy to make it a more easily-installed depdendency. @ewu63 have you had a chance to start that?

I have been away for a bit but will get back to this next week. I had started something which compiled fine locally but had trouble running it within the conda-forge docker container ecosystem, where some of the variables were not being passed in correctly for cmake to find library paths for blas etc.

**Comment by A-CGray** [comment]
Tests will fail until we merge the PR updating our docker images to unopy 0.4.7

**Comment by cvanaret** [comment]
> Thanks @cvanaret , I've been working on getting that major iteration tracking working with that callback. I have it working but there seems to have been some change between unopy v0.4.6 and 0.4.7 that means Uno can no longer converge the rosenbrock test problem. With 0.4.6 it converges in 47 major iterations, whereas with 0.4.7 I killed the process after about 1500 major iterations without convergence. I've attached two log files from the two versions, where the 0.4.7 run was limited to 100 major iterations, I don't see any obvious differences in the algorithm settings.
> 
> [Uno 046-rosenbrockOutput.log](https://github.com/user-attachments/files/27781294/Uno.046-rosenbrockOutput.log) [Uno 047-rosenbrockOutput.log](https://github.com/user-attachments/files/27781295/Uno.047-rosenbrockOutput.log)

Ah, I changed the [initialization of L-BFGS](https://github.com/cvanaret/Uno/pull/666) between 0.4.6 and 0.4.7. This is most likely the culprit (no wonder, it's got the number 666 :smiling_imp: ). I will revert. I will also add a CI workflow that runs the pyOptSparse benchmark with the latest Uno changes, to make sure unopy doesn't regress.

**Comment by cvanaret** [comment]
Fixed in unopy v0.4.8!

**Comment by A-CGray** [comment]
I just merged a PR that bumps our docker images up to unopy 0.4.8, usually takes ~4 hours to run all the CI tests and push the new images, at which point the tests here should run successfully.

File: tests/test_hs015.py [hunk 9]
```diff
@@ -127,6 +127,72 @@ def test_optimization(self, optName):
...
+        self.setup_optProb()
+        # Set loose convergence criteria
+        optOptions = {
+            "primal_tolerance": 1e-3,
+            "dual_tolerance": 1e-3,
+        }
+        sol = self.optimize(optOptions=optOptions)
+        # Check that solution is found (though not as tight as default)
+        self.assert_inform_equal(sol)
+
```
A-CGray: Is this supposed to test that we can successfully change the Uno convergence tolerances through the options? If so then I don't think testing that we converge to the same solution is a good approach because that's exactly what we'd expect to happen if we weren't able to change the convergence criteria.

tests/test_hs015.py
```diff
@@ -127,6 +127,72 @@ def test_optimization(self, optName):
...
+    def test_uno(self):
+        self.optName = "Uno"
+        self.setup_optProb()
+        sol = self.optimize()
+        # Check Solution
+        self.assert_solution_allclose(sol, 1e-4)
+        # Check informs
+        self.assert_inform_equal(sol)
+
+    @parameterized.expand(["filtersqp", "filterslp", "funnelsqp"])
```
A-CGray: Why not test the `ipopt` preset too? Or is that covered in `test_uno`?

A-CGray (changes requested):

tests/test_hs015.py
robfalck: I think initially ipopt wasn't working, but updates to Uno should have fixed that.  Will fix

ewu63:
Just a heads up that I have managed to get unopy working in conda-forge after some back and forth. PR is [here](https://github.com/conda-forge/staged-recipes/pull/33412) and there will be some adjustments + potential porting of patches upstream. We can coordinate the activities in that PR, and feel free to add yourselves as maintainers.

pyoptsparse/pyUno/pyUno.py
```diff
@@ -0,0 +1,458 @@
...
+        self,
+        optProb,
+        sens=None,
+        sensStep=None,
+        sensMode=None,
+        storeHistory=None,
+        hotStart=None,
+        storeSens=True,
+    ):
+        """
```
cvanaret, line 165: The `__call__` documentation should be removed/updated.

A-CGray (dismissed):

A-CGray (approved):
Thanks for all your work on this @robfalck and @cvanaret , this is a great addition to pyOptSparse!

pyoptsparse/pyUno/pyUno.py
marcomangano: What is wrong with the current docstrings?

cvanaret: Ah I thought it was copied-pasted from some other unrelated solver. But it looks solver-independent, my bad!

marcomangano (approved):
Great work, thanks a lot!

robfalck:
Thanks for the continued support for pyoptsparse. It really is a valuable tool. And @cvanaret, thanks for an open-source tool as capable as Uno.

It's been a pleasure collaborating with all of you on this.

cvanaret:
Thank you all for your time! I can't wait to improve the existing Uno solvers and try out new ones.

PR #486, opened 2026-06-16
Add sparse matrix and scaling tests

ewu63:
## Purpose
Add a few tests, partially addresses #256.

## Expected time until merged
A few days

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


tests/test_gradient.py
```diff
@@ -0,0 +1,88 @@
...
+        optProb = build_optProb()
+        funcs, _ = objfunc(X0)
+        grad = Gradient(optProb, sensType=sensType)
+        funcsSens, fail = grad(X0, funcs)
+        self.assertFalse(fail)
+        atol = 1e-12 if sensType == "cs" else 1e-5
+        assert_sens_matches_analytic(funcsSens, atol=atol)
+
+        # test that we get real derivs for cs
+        if sensType == "cs":
```
marcomangano, line 66: This is interesting, I would have thought that the `assert_allclose` would have caught a type mismatch. Would it make sense to move this before the call to `assert_sens_matches_analytic`?

tests/test_gradient.py
```diff
@@ -0,0 +1,88 @@
...
+    c   = x0*x1 + y
+    """
+    x, y = xdict["x"], xdict["y"]
+    funcs = {}
+    funcs["obj"] = x[0] ** 2 + 2 * x[1] ** 2 + 3 * y**2
+    funcs["c"] = x[0] * x[1] + y
+    return funcs, False
+
+
+def build_optProb(objfun=objfunc, xScale=1.0, conScale=1.0):
```
marcomangano, line 35: Minor curiosity: Is the mismatch between arg and function name intentional? Naming best practice?

tests/test_gradient.py
```diff
@@ -0,0 +1,88 @@
...
+        atol = 1e-12 if sensType == "cs" else 1e-5
+        assert_sens_matches_analytic(funcsSens, atol=atol)
+
+        # test that we get real derivs for cs
+        if sensType == "cs":
+            for funcKey in ANALYTIC:
+                for dvGroup in ANALYTIC[funcKey]:
+                    self.assertFalse(np.iscomplexobj(funcsSens[funcKey][dvGroup]))
+
+    def test_failed_eval(self):
```
marcomangano, line 71: For sake of future maintainability, do you mind adding a couple of lines of comment to describe these last two tests?
The first one could be "Testing that failed flags from objfunc are caught by the Gradient class", and TBH I am confused about what we are testing in the last one. Is that to ensure that the original gradient object is not scaled, similar to l.350 in `test_scaling.py`?

tests/test_optProb.py
```diff
@@ -295,5 +296,123 @@ def test_parallel_add(self):
...
+        assert_allclose(gobj, gobj_orig)
+
+    def test_mapConJac_formula_and_roundtrip(self):
+        # Build an arbitrary dense Jacobian of the right shape and convert to CSR.
+        rng = np.random.default_rng(2)
+        dense = rng.uniform(-2, 2, (self.nCon, self.ndvs))
+        jac = convertToCSR(dense)
+
+        # _mapConJactoOpt works in place: J_opt = diag(conScale) . J . diag(invXScale)
+        self.optProb._mapConJactoOpt(jac)
```
marcomangano, line 371: This is well outside the scope of the PR, but I wonder what is the rationale for `_mapObjGradtoOpt` returning an object (operating on a copy of the original `gobj`), while `_mapConJactoOpt` works in place?

tests/test_utils.py
```diff
@@ -0,0 +1,254 @@
...
+        rows, cols, data = coo["coo"]
+        assert_array_equal(rows, [0, 0, 1, 2, 2])
+        assert_array_equal(cols, [0, 2, 1, 2, 0])
+        assert_allclose(data, [1.0, 2.0, 3.0, 5.0, 4.0])
+        self.assertEqual(coo["shape"], [3, 3])
+
+    def test_from_csc(self):
+        coo = convertToCOO(_CSC)
+        rows, cols, data = coo["coo"]
+        assert_array_equal(rows, [0, 2, 1, 0, 2])
```
marcomangano, line 112: These are hardcoded and different from `_COO` because the "reading order" changes based on the initial format, right? Could you add a comment line similarly to what you did in the following tests?

marcomangano (commented):
Thanks a lot for doing this! Just a few very minor points below

tests/test_gradient.py
ewu63: A little bit? objfunc is what we use to call functions, but Optimization object's field is objfun. I can harmonize this.

tests/test_gradient.py
ewu63: `strict=True` would assert dtype also but defaults to False. Typically we want it to broadcast and ignore shape but not dtype.

tests/test_optProb.py
ewu63: It's a bit of an implementation detail. `_mapObjGradtoOpt` operates on a numpy array directly, while `_mapConJactoOpt` is from older code that also operates on sparse matrices, and for efficiency it mutates in place. This is something we should probably unify, and i think some of the type hints here are wrong.

tests/test_gradient.py
ewu63: Yes the difference here is using CS, rather than analytic derivatives. This is more integration test while unit tests for the mapping funcs are below.

tests/test_gradient.py
ewu63: Oh no I remember now, I didn't want to shadow a function with a variable with the same name.

tests/test_utils.py
ewu63: Addedd

marcomangano (dismissed):
Thanks for the comments, LGTM

tests/test_optProb.py
```diff
@@ -295,5 +296,123 @@ def test_parallel_add(self):
...
+        x_user = [3.0, 7.0]  # note x_user[0] == offset
+        x_opt = optProb._mapXtoOpt(x_user)
+        # (x - 3) * 4
+        assert_allclose(x_opt, [0.0, 16.0])
+        assert_allclose(optProb._mapXtoUser(x_opt), x_user)
+
+    def test_infinite_bounds_not_scaled(self):
+        """INFINITY bounds must remain unbounded; scale/offset must not turn
+        them into finite numbers in the assembled bounds.
+        """
```
A-CGray, line 419: Can you add docstrings like this, that explain what the test is checking, to all these new tests? I usually find LLMs are pretty good at writing them, especially if you tell them to use the [Given-When-Then](https://martinfowler.com/bliki/GivenWhenThen.html) style.

A-CGray (changes requested):

tests/test_optProb.py
ewu63: Added. I didn't exactly follow the Gherkin format but should be enough to follow.

A-CGray: Thanks!

A-CGray (approved):

PR #485, opened 2026-06-15
Add Egor optimizer wrapper

relf:
## Purpose

This PR integrates the surrogate-based optimizer `Egor` from the [EGObox](https://github.com/relf/EGObox) library.

Egor uses [bayesian optimization](https://en.wikipedia.org/wiki/Bayesian_optimization) techniques well-suited to find the global optimum of an expansive-to-evaluate black-box function.
Basically, it uses a surrogate model to approximate the objective function and an infill criterion (aka acquisition function) to guide the search for the optimum.

Resolves https://github.com/relf/EGObox/issues/439

## Expected time until merged

a few weeks

## Type of change

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [x] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Egor tests have been added to the pyOptSparse test suite.

## Checklist

- [x] I have run `ruff check` and `ruff format` to make sure the Python code adheres to PEP-8 and is consistently formatted
- [ ] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable
- [x] I have run unit and regression tests which pass locally with my changes
- [x] I have added new tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation

PR #487, opened 2026-06-16
Add optimization test with FD derivatives

ewu63:
## Purpose
Addresses part of #256.

## Expected time until merged
A few days.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


marcomangano:
Which one should be merged first between this PR and  #486 ? Both add the same two new files, does it make sense attach them to one PR over the other?

ewu63:
This is on top of #486, let's wait for that one first.

pyoptsparse/pyOpt_gradient.py
```diff
@@ -44,6 +44,12 @@ def __init__(self, optProb: Optimization, sensType: str, sensStep: float = None,
                 self.sensStep = 1e-40j
         else:
             self.sensStep = sensStep
+
+        # Complex step divides by the imaginary part of the step, so a purely
+        # real step would silently yield NaN gradients.
+        if self.sensType == "cs" and np.imag(self.sensStep) == 0:
+            raise ValueError(f"The complex step size must have a nonzero imaginary part, got {self.sensStep}.")
```
A-CGray: The docstring says that `sensStep` is the step `size`, and should be a `float`. So we need to either

1. Update the docstring
2. Change the implementation so that we take in a float and do `self.sensStep = 1j * sensStep` when using complex-step

Option 2 would be my slight preference.

A-CGray: Also, if we keep the implementation as is and update the docstring, shouldn't we also check that `sensStep` is purely imaginary (i.e `np.real(self.sensStep) == 0`?)

A-CGray (changes requested):
Same overall comment as #486, can you add docstrings for new tests explaining what they are testing?

Also, is there a preferred order in which we should merge this and #486 ?

pyoptsparse/pyOpt_gradient.py
ewu63: I would prefer not to change any code behaviour (other than additional error checking as done here). The "correct" path can be updated later on if we want to, but definitely require a separate PR and more documentation. I will update the docstring to be more clear, but yes as coded the user needs to pass the actual complex perturbation. In practice, the default of 1e-40j should work for almost all scenarios.

ewu63:
This PR should be merged after #486, I will rebase it after that one is merged.

tests/test_sphere.py
```diff
@@ -101,6 +105,17 @@ def test_optimization(self, optName):
         optOptions = self.optOptions.get(optName, {})
         self.optimize_with_hotstart(self.tol[optName], optOptions=optOptions)
 
+    @parameterized.expand(
+        product(ALL_OPTIMIZERS, ["fd", "fdr", "cd", "cdr", "cs"]),
+        name_func=lambda f, n, p: f"{f.__name__}_{p.args[0]}_{p.args[1]}",
+    )
+    def test_optimization_approx_deriv(self, optName, sens):
```
ewu63, line 112: This is the only new test added here, the rest are from the other branch (the diff will look better once the other PR is merged).

A-CGray (dismissed):

ewu63:
Sorry had to rebase `--onto` since we squash PRs. Need another approval @A-CGray. I plan to merge #485 after this one and make a new release.

A-CGray (dismissed):

A-CGray:
@marcomangano

pyoptsparse/testing/pyOpt_testing.py
```diff
@@ -356,46 +356,42 @@ def check_hist_file(self, tol):
             for varName in self.DVs:
                 assert_allclose(val[varName].flatten(), self.xStar[self.sol_index][varName], atol=tol, rtol=tol)
 
-    def optimize_with_hotstart(self, tol, optOptions=None, x0=None):
+    def optimize_with_hotstart(self, tol, x0=None, **kwargs):
```
marcomangano: Is replacing `optOptions` with the general `**kwargs` in the function API a syle/formatting upgrade or is there another reason behind it?

marcomangano (commented):
The tests look good (see my question below), but as I was checking the whole file I noticed that  `self.sens` (which provides the analytical derivatives) is never used. Am I reading it wrong? Should we fix `test_optimization` to address that? Otherwise some of the `fd` tests might be partially overlapping (the new tests do not use `optimize_with_restart`)

pyoptsparse/testing/pyOpt_testing.py
ewu63: Huh, maybe the rebase got messed up or something, this is not necessary I can revert.

ewu63:
> The tests look good (see my question below), but as I was checking the whole file I noticed that `self.sens` (which provides the analytical derivatives) is never used. Am I reading it wrong? Should we fix `test_optimization` to address that? Otherwise some of the `fd` tests might be partially overlapping (the new tests do not use `optimize_with_restart`)

Not sure I'm following. This PR adds one main test `test_optimization_approx_deriv`, which runs the sphere unconstrained problem using approximate gradients (FD/CS etc.), so we do not touch `self.sens`. Gradient tests are already done in `test_gradient.py`, here we are doing integration tests with different gradient approx methods on a full optimization problem.

pyoptsparse/testing/pyOpt_testing.py
ewu63: Reverted

A-CGray (approved):

marcomangano:
Nevermind @ewu63 , I missed [this snippet](https://github.com/mdolab/pyoptsparse/blob/285c47d1d778cf6c216cc5a9cfafc1a0231cfdf9/pyoptsparse/testing/pyOpt_testing.py#L233-L235) in the parent `optimize` method that actually calls `self.sens`.

I was confused as it looked that the original `test_optimization` was also defaulting to FD derivatives. All good, thanks for adding this test!

marcomangano (approved):

PR #493, opened 2026-09-09
Fix SLSQP Lagrange multiplier extraction

ewu63:
## Purpose
SLSQP Lagrange multiplier was broken previously, it is fixed here and turns out to be easier than I thought.

## Expected time until merged

Would be nice to sneak into this upcoming release.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Added to existing hs071 test.

## Checklist


pyoptsparse/pySLSQP/pySLSQP.py
```diff
@@ -230,24 +230,13 @@ def slgrad(m, me, la, n, f, g, df, dg, x):
...
-                    else:
-                        lambdaStar.append(lambdaStar_upper)
-                    idx += 1
+            # Per the SLSQP docstring, on return W(1) ... W(M) contain the multipliers
+            # associated with the M general (one-sided) constraints, one entry per
+            # constraint, in the same order as self.optProb.jacIndices (set above from
+            # getOrdering). slfunc/slgrad negate fcon/gcon before handing them to SLSQP
+            # (since SLSQP expects constraints of the form c(x) >= 0), so the returned
+            # multipliers must be negated to match pyOptSparse's constraint sign convention.
+            lambdaStar = -w[:m]
```
marcomangano, line 239: Should we keep this consistent with the original notation and just flip the sign for testing purposes, like we do for IPOPT [here](https://github.com/mdolab/pyoptsparse/blob/263ecd64a74ff4e02fd001ac57ec713138519736/tests/test_hs071.py#L146)?

ewu63: I updated the comment to explain why, in particular this convention is consistent with scipy's SLSQP and in any case I want to move towards using scipy's implementation in the future. I am also slowly leaning towards flipping the sign of IPOPT's lagrange multiplier so that all optimizers when wrapped by pyOptSparse has a consistent sign convention, but we can discuss that later.

marcomangano (approved):
Happy to talk about IPOPT multipliers sign offline. LGTM

PR #494, opened 2026-09-23
Fix `getOrdering` equality expansion

ewu63:
## Purpose
There was a bug in the code that converts a single two-sided constraint to two one-sided constraints, for optimizers that cannot handle two-sided constraints directly (CONMIN, Egor, NSGA2). This meant that for these optimizers, optimization problems with nonlinear equality constraints where the constraint value is not zero could result in erroneous optimization solution. This was reproduced on the hs060 problem (added here) with Egor, where without the bugfix the solution was reported as [0.35, 0.34, 0.35] instead of [1, 1, 1].

## Expected time until merged
A few days.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Added unit and regression tests.

## Checklist


pyoptsparse/pyOpt_optimization.py
```diff
@@ -971,13 +971,13 @@ def getOrdering(
...
-                        lower.extend(econ["value"])
+                        lower.extend([-INFINITY] * len(econ["fact"]))
                         upper.extend(econ["value"])
                         # ....And the other side
                         indices.extend(con.rs + econ["ind"])
                         fact.extend(-1.0 * econ["fact"])
-                        lower.extend(econ["value"])
-                        upper.extend(econ["value"])
+                        lower.extend([-INFINITY] * len(econ["fact"]))
+                        upper.extend(-econ["value"])
```
marcomangano: Took me a second to figure the math out because I missed the change in sign of the factor. Should we add a short comment to explain? If I get this right:
`a=b ->  a<=b && a>=b  -> a<=b && -a=<-b`

marcomangano (commented):
Is the NLPQLP test failure a matter of convergence?
Also, see my comment below

